Issue #7, opened 2018-10-10
CMake support


FlorianRhiem:
Hello and thank you for your work!
Currently there are no files for any build systems aside from the shell script calling g++ that you provide with the examples. I think it'd be nice to have a CMakeLists.txt along with the source code to build and install the library, run the tests, etc. Would that be alright for you, or is the lack of something like that based on the project's zero dependency principle? If you simply don't use CMake yourself, then I could prepare a pull request during the weekend.

lganzzzo:
Hello @FlorianRhiem , thank you for your interest and readiness to help!
All infrastructural pull-requests that make devs life easier are definitely very welcome!

FlorianRhiem:
Good to hear! I just started setting something up and noticed that the way you include your files, you require your users to clone the repo or download the source as `oatpp`. Any other directory name will lead to failed includes. GitHub for example will give users an `oatpp-master` directory if they download the source as zip file, so this would need to be manually renamed.
Was the omission of that directory intentional? If not, I would suggest moving all files that are expected to exist in an `oatpp` directory to that directory. The big downside of this, obviously is that it isn't compatible with the current examples.

lganzzzo:
The way it is right now is intentional.

Current schema of ```includes``` looks as follows:

```Include``` paths are treated as ```"<module>/path/to/file"```
Currently ```oatpp``` has multiple modules (git submodules) to be included on demand.

- ```oatpp``` (basic framework) - no dependency
- ```oatpp-swagger``` (swagger-ui) - no dependency except oatpp itself
- ```oatpp-consul``` (Consul client - see consul.io) - no dependency except oatpp itself
- ```oatpp-libressl``` (TLS based on ```libressl```) - libressl required

To include new submodule:
in lib folder
```$ git submodule add <url>```

The idea to ship ```oatpp``` as full source-code (as git submodules) is to enable user:
- To easily debug everything
- Make process of contribution easy and natural (you found the bug you fixed it, you pushed the PR)

Such structure of the project is not final, and in some sens is an experiment.

FlorianRhiem:
Alright, I see the benefit in submodule use. I'll just use the parent as include directory for building then.

---

The three `parser::json::mapping` tests fail with uncaught exceptions. Are these still a work-in-progress or am I missing a required part of test setup?
```C++
#include "oatpp/test/parser/json/mapping/DeserializerTest.hpp"
#include "oatpp/test/parser/json/mapping/DTOMapperPerfTest.hpp"
[...]

PR #8, opened 2018-10-11
Use name pointer for internal type identification

FlorianRhiem:
This is a possible solution to the problem with `CLASS_NAME`s in shared libraries discussed in #7.  Types have the string as a non-`constexpr` static member and have a pointer to that static member as the `constexpr` that is used in comparisons etc. This way you have both a human-readable representation and a `constexpr`, with the difference that it must work thanks to the one definition rule.

A few things you might want to do differently to this branch:
- I have added the `namePtr` to `Type` and kept the `name` for the sake of simplicity. If you don't want to waste the memory of keeping `name` around, it could be a method that simply dereferences `namePtr`.
- I have kept the name `CLASS_NAME` and called the non-`constexpr` string `CLASS_NAME_NONCONSTEXPR`. `CLASS_NAMEPTR` and `CLASS_NAME` might be better names, but they would require larger changes to the existing code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,267 @@
+cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
+
+file(STRINGS "${CMAKE_CURRENT_LIST_DIR}/core/base/Environment.hpp" OATPP_VERSION_MACRO REGEX "#define OATPP_VERSION \"[0-9]+.[0-9]+.[0-9]+\"$")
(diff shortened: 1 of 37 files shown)


lganzzzo:
please change target branch to
[data_mapping_type_id_refactoring](https://github.com/oatpp/oatpp/tree/data_mapping_type_id_refactoring)

lganzzzo:
Fixed in #17

PR #10, opened 2018-10-12
CMake support

### LINKED ISSUE #7: CMake support

FlorianRhiem:
This PR adds a CMakeLists.txt to the project:
- The list file and possibly installed config files provide a library `oatpp::oatpp`.
- If `BUILD_SHARED_LIBS` is set, a shared library will be built. Otherwise (the default) a static library will be built.
- If `OATPP_INSTALL` is set (the default), an install target is generated that will install the library, its headers and cmake config files to the `CMAKE_INSTALL_PREFIX`.
- If `OATPP_BUILD_TESTS` is set (the default), the tests are built into a single executable and a test target is generated that will run these tests using CTest. To allow this, `test/AllTestsMain.cpp` was added that will provide a `main()` for running all tests if the macro `OATPP_ENABLE_ALL_TESTS_MAIN` is defined during compilation.
- To maintain CMakeLists.txt, new source and header files need to be added to the list of files for the `oatpp` library, or the `oatppAllTests` executable.

Fixes #7.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,267 @@
+cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
+
+file(STRINGS "${CMAKE_CURRENT_LIST_DIR}/core/base/Environment.hpp" OATPP_VERSION_MACRO REGEX "#define OATPP_VERSION \"[0-9]+.[0-9]+.[0-9]+\"$")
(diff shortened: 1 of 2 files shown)


lganzzzo:
That is a great PR!
Thank you for your help and interest in the project!

PR #33, opened 2019-01-18
Connection upgrade mechanism

lganzzzo:


core/data/stream/Stream.cpp
```diff
@@ -241,22 +241,47 @@ oatpp::async::Action transferAsync(oatpp::async::AbstractCoroutine* parentCorout
...
-oatpp::async::Action writeDataAsyncInline(oatpp::data::stream::OutputStream* stream,
-                                          const void*& data,
-                                          os::io::Library::v_size& size,
-                                          const oatpp::async::Action& nextAction) {
+oatpp::async::Action writeSomeDataAsyncInline(oatpp::data::stream::OutputStream* stream,
+                                              const void*& data,
+                                              os::io::Library::v_size& size,
+                                              const oatpp::async::Action& nextAction) {
   auto res = stream->write(data, size);
   if(res == oatpp::data::stream::Errors::ERROR_IO_WAIT_RETRY) {
```
Sergii-Ladniuk: Is there a way to avoid repeating things like "oatpp::data::stream::Errors::" ?

core/data/stream/Stream.cpp
```diff
@@ -241,22 +241,47 @@ oatpp::async::Action transferAsync(oatpp::async::AbstractCoroutine* parentCorout
...
   } else if(res == oatpp::data::stream::Errors::ERROR_IO_PIPE) {
     return oatpp::async::Action::_ABORT;
-  } else if( res < 0) {
+  } else if(res < 0) {
     return oatpp::async::Action(oatpp::async::Error(Errors::ERROR_ASYNC_FAILED_TO_WRITE_DATA));
-  } else if(res < size) {
-    data = &((p_char8) data)[res];
-    size = size - res;
+  }
+  data = &((p_char8) data)[res];
```
Sergii-Ladniuk: it's hard to understand what's going on, maybe better split

core/data/stream/Stream.cpp
```diff
@@ -241,22 +241,47 @@ oatpp::async::Action transferAsync(oatpp::async::AbstractCoroutine* parentCorout
   
 }
   
-oatpp::async::Action writeDataAsyncInline(oatpp::data::stream::OutputStream* stream,
-                                          const void*& data,
-                                          os::io::Library::v_size& size,
-                                          const oatpp::async::Action& nextAction) {
+oatpp::async::Action writeSomeDataAsyncInline(oatpp::data::stream::OutputStream* stream,
+                                              const void*& data,
```
Sergii-Ladniuk: why do you pass a reference to pointer?

core/data/stream/Stream.cpp
lganzzzo: Yes, one may use namespace aliases, or `using` in user's code.
But in this exact place it's redundant. I'll remove that

PR #75, opened 2019-05-21
unique_ptr for IOEventWorker::m_inEvents

lganzzzo:


src/oatpp/core/async/worker/IOEventWorker_common.cpp
```diff
@@ -40,16 +40,14 @@ IOEventWorker::IOEventWorker(IOEventWorkerForeman* foreman, Action::IOEventType
   , m_wakeupTrigger(-1)
   , m_inEvents(nullptr)
```
elfring: Can this explicit initialisation be omitted because of the smart pointer usage?

PR #78, opened 2019-05-31
Fix compiler errors and warnings in Android and iOS

dmcamens-legrand:
This pull request removes compiler errors and warnings that we found when building oatpp for Android and iOS.


lganzzzo:
Merged. 
Thank you guys!

PR #79, opened 2019-05-31
Set thread affinity on linux but not on android

lganzzzo:


lganzzzo:
@dmcamens-legrand , @apple-ihack-geek 
Please check out this PR

src/oatpp/core/concurrency/Thread.cpp
```diff
@@ -36,27 +36,36 @@ v_int32 setThreadAffinityToOneCpu(std::thread::native_handle_type nativeHandle,
...
   // The below line compiles on Android but has not been tested.
   //result = sched_setaffinity(nativeHandle, sizeof(cpu_set_t), &cpuset);
-  
-  if (result != 0) {
-    OATPP_LOGD("[oatpp::concurrency::Thread::assignThreadToCpu(...)]", "error code - %d", result);
-  }
-  
-  return result;
+
+  #ifndef __ANDROID__
```
dmcamens-legrand: Maybe combine this line with the above "#if defined(_GNU_SOURCE)"?  This would allow you to remove the redundant "#else return -1" at the bottom of the method.

lganzzzo: Yes you are right. Ideally would be add Android-specific functionality also

PR #105, opened 2019-08-05
Better Windows-defineswitches and better usage of winsock api

bhorn:
This PR fixes some typing-issues in Windows and optimizes all ifdefs for windows.
Further, the winsockapi is better incorporated to further reduce technical debt. 
Virtual tests still run fine.


lganzzzo:
Nice!

lganzzzo:
Merged

PR #109, opened 2019-08-12
Warningless in -Wextra, hardened sprintf-calls.

bhorn:
Our code policy requires us to not use `sprintf` (only `snprintf` allowed) as well to write warningless code with `-Wextra` (and listen to clang-tidy as much as possible). I thought this may be relevant for code-quality/hardening of oatpp-upstream as well.


lganzzzo:
Merged!
Great job!

PR #112, opened 2019-08-18
Refactor coroutine member caller

lganzzzo:


lganzzzo:
Successful CI check - https://dev.azure.com/lganzzzo/lganzzzo/_build/results?buildId=2359

portsip:
Seems the performance is not as well as Linux edition ?

lganzzzo:
Hey @portsip ,

My "theory" is that performance should be pretty similar among platforms - at least for **non-async** oatpp servers. However I don't have any cross-platform performance tests conducted - so that the same application would be ran at different platforms but with the same computing powers.

---

### Do not look at CI for performance comparison 

At CI you may see that Windows build takes about twice more time than the linux build.
But CI is not for performance testing and its unknown what machines (in terms of computing powers) were scheduled to run Linux/Unix/Windows builds.

Also if we look at CI-test `web::FullTest` on "real port"(8000):

|OS|Time(microseconds)|
|---|---|
|Linux|9,806,891|
|Mac|1,251,490|
|Windows|1,686,795|

Windows performance for this particular test is 6x times better than Linux performance.
**But again - CI is not for performance testing.**

---

If you will, or have already conducted any performance testing on Windows please share your results - it will be very interesting!

Best Regards,
Leonid

PR #113, opened 2019-08-21
Link against libatomic

bhorn:
In some environments, libatomic is not part of threads/stdlib/whatsoever. So for these environments, the libatomic needs specific linking. This is (for example) the case in some armvX and mips environments or openwrt.


lganzzzo:
Hey @bhorn ,

Thanks for the PR!

Merged.

PR #114, opened 2019-08-29
Authorization

bhorn:
This PR adds a new `AUTHORIZATION` macro to the list of available endpoint-macros.
The new macro passes the `Authorization` header to the newly introduced `oatpp::web::server::handler::AuthorizationHandler`. A dedicated `AuthorizationHandler` can be registered for each `ApiController`. If no `AuthorizationHandler` is registered, the `ApiController` will call `DefaultAuthorizationHandler::handleDefaultAuthorization` which returns an `DefaultAuthorizationObject` containing the pares username and password of the request.
One can implement his own `AuthorizationHandler` and `AuthorizationObject` so its easy to implement database-driven authentication and yield an object reflecting the authenticated user with all needed data, so no extra work in the endpoint-code needs to be done.

See https://github.com/bhorn/oatpp-authorization-example for an example covering all possible cases on how to use the new mechanisms.


lganzzzo:
Looks good!

PR #128, opened 2019-09-26
 Fix. Fixed crash happening when debuggng via GDB.

bkannadassan:
Problem: Crash seen when loading build with oatpp.

Root Cause: When loading the build in gdb process gets EINTR and eventsCount is -1, hence crashing.

Fix: Its good to check EINTR before throwing error, hence added the same..

rgds
Balaji Kamal Kannadassan


lganzzzo:
Merged! Great job!

PR #139, opened 2019-10-11
 Two Fixes with CORS and Swagger, one addition to `oatpp::String` to accomplish one fix

bhorn:
- **Fix controller-router-prefix messing up CORS OPTIONS endpoints:**
    Prefix would be prepended twice!
- **Fix missing path-params in swagger for CORS OPTIONS endpoints:**
    Confuses most code-generators, however only relevant if CORS OPTIONS are not hidden
- **Added substr-methods to `oatpp::StrBuffer` for convenience and easier implementation of the double-prefix-bug**


src/oatpp/core/base/StrBuffer.hpp
```diff
@@ -247,6 +247,21 @@ class StrBuffer : public oatpp::base::Countable {
    * @return - true if buffer starts with specified data.
    */
   bool startsWith(StrBuffer* data) const;
+
+  /**
+   * Retuns a copy of the substring starting at the given index.
+   * @param start - index from where to start
+   * @return - copy of the substring
+   */
+  std::shared_ptr<StrBuffer> substr(v_int32 start);
```
lganzzzo: Please remove newly added API for `StrBuffer` as its not used in the fix.

lganzzzo:
Merged!

PR #148, opened 2019-11-01
Clean some warnings in Clang 10/Xcode 10.3

rwtolbert:
Don't use std::move when it prevents copy elision (RVO?)
Reorder some initialization statements. 
and remove an unused variable.


lganzzzo:
Looks good!

PR #155, opened 2019-11-22
Move string to int64

rwtolbert:
Move String and related data types and sizes to v_int64.

Consistent size everywhere and removes tons of conversion warnings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ cmake-build-debug/
 
 # local build
 build/
+build-xcode/
 srt/build/
 test/build/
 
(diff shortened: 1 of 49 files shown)


bhorn:
Just as comment, even if its already merged. I would propose to use something like v_size.
Hardcoding v_int64 for every size-type can be a (huge) performance penalty for native 32bit applications like in small uCs. 
IE for MIPS where there are no overflow-flags and other indicators for arithmetics, handling 64 bit arithmetic is a huge performance penalty since the whole 64bit handling is software defined with a lot of checks, branches and calculations.
Therefore using v_size and make it platform bitness by default (and maybe overrideable with an define switch) would be a much more versatile solution and help oatpp's compatibility and performance on small devices.

lganzzzo:
Please note - changes introduced by this PR were merged as part of #156 .
Sorry for the confusion.

PR #156, opened 2019-11-23
Move string to int64

lganzzzo:


lganzzzo:
Hello @bhorn ,
Please note that changes introduced by #155 , were merged as part of this PR.

lganzzzo:
Hey @bhorn ,
The summary of this PR:

- Introduced [v_buff_size](https://github.com/oatpp/oatpp/blob/master/src/oatpp/core/base/Environment.hpp#L77) data type as an integer capable of storing the pointer.
- Moved String and buffers to v_buff_size for their sizes.

So v_int64 is not hardcoded for all buffer-sizes. And it can be easily changed to whatever other types.

Regards,
Leonid

PR #160, opened 2019-12-08
Builds on UWP

mheyman:
Hex.cpp and Unicode.cpp were including winsock.h instead of winsock2.h which caused undefined functions for Universal Windows Platform builds.


lganzzzo:
Merged! 
Great job!

PR #183, opened 2020-01-15
Add INET support for linux (not just INET6)

ThmX:
Gives the ability to use oat++ on an embedded system without IPv6.


lganzzzo:
Awesome! 
Merged!

bhorn:
Hey, may I ask you which embedded Linux you are using on which target CPU/SoC/SoM?

ThmX:
> Hey, may I ask you which embedded Linux you are using on which target CPU/SoC/SoM?

i.MX6 UltraLite 2 (Cortex-A7 528MHz, 256 MB DDR) running Linux built with OpenEmbedded

PR #186, opened 2020-01-19
fix crashed

gergul:
`multipartReader->setPartReader("file1", multipart::createAsyncFilePartReader("/path/to/file")); `
When "/path/to/file" is a invalid path. It will crashed.


lganzzzo:
Merged! 🎉

Issue #191, opened 2020-01-28
Issues building on FreeBSD 12.1


abenedic:
I had an issue compiling on FreeBSD 12.1 due to a missing header include. Adding:

    #if defined(__FreeBSD__) 
        #include <netinet/in.h>
    #endif

to the includes for non-windows systems in oatpp/src/oatpp/network/server/SimpleTCPConnectionProvider.cpp resolves the issue.  

Another issue came during linking since it links with -latomic which does not work on FreeBSD as -latomic is not needed.  Adding: 

    if(MSVC)
        target_link_libraries(oatpp PUBLIC ${CMAKE_THREAD_LIBS_INIT} wsock32 ws2_32)
    elseif(APPLE)
        target_link_libraries(oatpp PUBLIC ${CMAKE_THREAD_LIBS_INIT})
    else(MSVC)
        if(${CMAKE_SYSTEM_NAME} STREQUAL "FreeBSD")
            target_link_libraries(oatpp PUBLIC ${CMAKE_THREAD_LIBS_INIT})
        else(${CMAKE_SYSTEM_NAME} STREQUAL "FreeBSD")
            target_link_libraries(oatpp PUBLIC ${CMAKE_THREAD_LIBS_INIT} atomic)
        endif(${CMAKE_SYSTEM_NAME} STREQUAL "FreeBSD")
    endif(MSVC)

to the CMakeLists.txt in the src directory resolved this for me, although it is ugly.

lganzzzo:
Merged!
Thanks for the fix!

PR #192, opened 2020-01-28
Minor changes for Building on FreeBSD 12.1

abenedic:
I don't really use CMake much so my solution to the linking issue is bad, but in case others aren't aware of the issue here is my fix.


lganzzzo:
Hello @abenedic ,

I think that the fix is good enough. It's specific and can be extended for other BSD OSs in case it should be.

Links:
- https://github.com/oatpp/oatpp/issues/191

Regards,
Leonid

lganzzzo:
Merged!

Issue #198, opened 2020-02-09
Server deserialize data very slowly when request body contains much data


BossZou:
I updated oatpp from v0.19.11 to v1.0.0, found that when I sent a `POST` request which body contained much data, the serve cost more time to do deserialization. 

Here is my DTO definition:
```
class InsertRequestDto : public oatpp::data::mapping::type::Object {
    DTO_INIT(InsertRequestDto, Object)

    DTO_FIELD(String, tag) = VALUE_PARTITION_TAG_DEFAULT;
    DTO_FIELD(List<List<Float32>::ObjectWrapper>::ObjectWrapper, records);
    DTO_FIELD(List<List<Int64>::ObjectWrapper>::ObjectWrapper, records_bin);
    DTO_FIELD(List<Int64>::ObjectWrapper, ids);
};
```

When the `records` is a 100000 * 128 array, the server will cost about 10s to deserialize it at version 0.19.11, but about 800s at v1.0.0. I noticed that the Deserializer has been optimized after v0.19.11. But it got too slow in my case.

lganzzzo:
Hey @BossZou ,

Thanks for the issue!

Do you have a sample json file that you can provide to us to test?

BossZou:
Here is my python code to generate json file:
```
import json
import random


if __name__ == '__main__':
    vectors = [[random.random() for _ in range(128)] for _ in range(100000)]

    data = {
        "vectors:": vectors
    }

    with open('result.json', 'w') as fp:
        json.dump(data, fp)
```
I'm sorry that the json file is bigger than 10MB. I can't upload it to github

lganzzzo:
Hey @BossZou ,

FYI, this issue is fixed in the `optimize_large_files_upload` branch.
It will be merged into master once PR is approved.

Regards,
Leonid

lganzzzo:
The fix was merged to master. 
Closing issue.

PR #200, opened 2020-02-16
Use stream::ChunkedBuffer instead of stream::BufferOutputStream for uploads

lganzzzo:
Use of `stream::BufferOutputStream` for uploads is inefficient due to excessive memory copy operations during buffer growth, taking too long to upload large files.

Using `stream::ChunkedBuffer` for uploads fixes this problem due to the no-copy buffer growth strategy.

Links:
- https://github.com/oatpp/oatpp/issues/198


bhorn:
Nice, noticeable difference (1-2s) on our small embedded systems even for 9mb.
Merged. Updating OpenWRT-Recipes, soon.

PR #204, opened 2020-03-24
Show http-method in logRouterMappings

bhorn:
`Router::logRouterMappings` never shows the http-method a URL is mapped for. So when you have multiple methods for one URL, the method will display the URL multiple times. This could confuse users why their URL is shown multiple times ("Have I added the URL twice?", "Do I call `logRouterMappings` multiple times?", etc). Its also not easily visible if all methods are mapped as designed.

This small PR adds the http-method to the output of `logRouterMappings`.


lganzzzo:
Looks good!

PR #208, opened 2020-03-29
Fix comparison of signed and unsigned integers.

hermanzdosilovic:
To comply with `-Wsign-compare` flag.

PR #209, opened 2020-03-30
Feature/fix sign compare

lganzzzo:


lganzzzo:
- Minor PR review changes to https://github.com/oatpp/oatpp/pull/208
- Fix unsigned int rounded in the loop
  /* bug. when m_processors.size() == 0, stop condition will be rounded to max int */
  for(v_uint32 i = 0; i < m_processors.size() - 1; i ++)  { 
     ...
  }

PR #214, opened 2020-04-03
Refactor. Remove confusing 'empty()' method of ObjectWrappers.

lganzzzo:
### Background

Users were confused by `oatpp::String::empty()` method which was returning just empty string instead of checking string for being empty.

---

The new `oatpp::String::empty()` method, which does the check of the string being empty 
**should NOT be introduced** in a couple of next releases - in order to NOT introduce unexpected behavior in the user's code.

### DIFF
--- a/src/oatpp/core/data/mapping/type/Primitive.hpp
+++ b/src/oatpp/core/data/mapping/type/Primitive.hpp
@@ -164,6 +164,10 @@ public:
     ObjectWrapper()
       : type::ObjectWrapper<Primitive, Clazz>()
     {}
+
+    ObjectWrapper(std::nullptr_t)
+      : type::ObjectWrapper<Primitive, Clazz>()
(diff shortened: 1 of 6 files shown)


bhorn:
Yes, first removing the old method before reintroducing is a great idea!

lganzzzo:
> Very good with switching to nullptr! However, don't you think that it would be nice to also do it for other primitives to keep a consistent code base?

Yep,  same for other primitives too - https://github.com/oatpp/oatpp/pull/214/files/646f22a543840c855fe26bc85077aa103ac0e7e5#diff-1b676e9171ef63bc3c0eff67c29271b1L217

`empty()` method is deleted for all the `ObjectWrappers` - for `Strings`, `Primitive<T>`, `ObjectWrapper`, `PolymorphicWrapper`

PR #220, opened 2020-04-11
Extend 'oatpp/core/Types.hpp' with List, Fields, Object.

lganzzzo:


bhorn:
Perfect! This will come in handy.

PR #223, opened 2020-04-20
Better float to string

lganzzzo:


bhorn:
Nice first step towards fixing our beloved localization issue!

PR #254, opened 2020-06-03
add check for gcc version less than 5 due to missing std::put_time

cdigruttola:
Hello,

I created this PR because I'm having an issue about GCC version less than 5. The problem is due to std::put_time (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54354#c9). I added also in the .gitignore some generated folders and file of Visual Studio.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,7 @@ test/build/
 
 Dockerfile
 
+# VS
+.vs/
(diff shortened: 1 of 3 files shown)


lganzzzo:
Merged!
Great Work!

PR #265, opened 2020-06-27
Fixed warnings in MSVS2019 with Warning level 4 (just oatpp static lib project)

srogatch:
I left some TODO: for @lganzzzo to consider because I don't have the competence in the project to fix that complex cases.


src/oatpp/core/async/worker/IOEventWorker_common.cpp
```diff
@@ -40,8 +40,8 @@ IOEventWorker::IOEventWorker(IOEventWorkerForeman* foreman, Action::IOEventType
   , m_foreman(foreman)
   , m_specialization(specialization)
   , m_running(true)
-  , m_eventQueueHandle(-1)
-  , m_wakeupTrigger(-1)
+  , m_eventQueueHandle(v_io_handle (-1))
```
lganzzzo, line 43: It might be a good idea to declare and `INVALID_IO_HANDLE` constant [here](https://github.com/oatpp/oatpp/blob/master/src/oatpp/core/IODefinitions.hpp#L36) together with `v_io_handle`.

Which is `-1` on Linux and `INVALID_SOCKET` on Windows.

src/oatpp/core/async/worker/IOWorker.cpp
```diff
@@ -78,8 +78,11 @@ void IOWorker::run() {
   v_int32 consumeIteration = 0;
   v_int32 roundIteration = 0;
 
-  std::chrono::microseconds ms = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch());
-  v_int64 tick = ms.count();
+  v_int64 tick;
+  {
```
lganzzzo, line 82: What if we just assign directly to `tick` without `ms` var.
```cpp
tick = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
```

Creating a separate block may look confusing.

src/oatpp/core/data/buffer/FIFOBuffer.cpp
```diff
@@ -362,7 +362,8 @@ v_io_size FIFOBuffer::flushToStream(data::stream::OutputStream* stream) {
   if(m_readPosition < m_writePosition) {
     result = stream->writeExactSizeDataSimple(&m_buffer[m_readPosition], m_writePosition - m_readPosition);
   } else {
-    auto result = stream->writeExactSizeDataSimple(&m_buffer[m_readPosition], m_bufferSize - m_readPosition);
+    //TODO: what was the original intent here? The statement before my fix had no effect on the result returned
```
lganzzzo, line 365: Yep, it looks like a bug!

src/oatpp/core/data/mapping/type/Any.cpp
```diff
@@ -38,7 +38,10 @@ Any::Any(std::nullptr_t) : Any() {}
 
 Any::Any(const std::shared_ptr<AnyHandle>& handle, const Type* const type)
   : ObjectWrapper(handle, __class::Any::getType())
-{}
+{
+  //TODO: is there indeed no use for |type| parameter passed above?
```
lganzzzo, line 42: As an `ObjectWrapper`, `Any` must have this constructor. It is used in `ObjectWrapper.staticCast<T>(...)` method.
However, unlike other `ObjectWrapper`s `Any` won't change its `valueType`.
`Any` is always object wrapper above `AnyHandler`.

src/oatpp/network/client/SimpleTCPConnectionProvider.cpp
```diff
@@ -75,7 +75,7 @@ std::shared_ptr<oatpp::data::stream::IOStream> SimpleTCPConnectionProvider::getC
   }
 
   struct addrinfo* currResult = result;
-  oatpp::v_io_handle clientHandle =- 1;
+  oatpp::v_io_handle clientHandle = oatpp::v_io_handle(-1);
```
lganzzzo, line 78: Let's use a new `INVALID_IO_HANDLE` constant here, from the previous comment.

src/oatpp/parser/json/Utils.cpp
```diff
@@ -203,52 +203,65 @@ oatpp::String Utils::escapeString(p_char8 data, v_buff_size size, bool copyAsOwn
...
-        resultData[pos] = '\\'; resultData[pos + 1] = 'r'; pos += 2;
-      } else if(a == '\t'){
-        resultData[pos] = '\\'; resultData[pos + 1] = 't'; pos += 2;
-      } else {
-        resultData[pos] = '\\';
-        resultData[pos + 1] = 'u';
-        oatpp::encoding::Hex::writeUInt16(a, &resultData[pos + 2]);
-        pos += 6;
+
+  {
```
lganzzzo, line 209: Just curious, why an additional block is used here?

src/oatpp/parser/json/mapping/Deserializer.hpp
```diff
@@ -133,6 +133,7 @@ class Deserializer {
     if(caret.isAtText("null", true)){
       return oatpp::Void(T::Class::getType());
     } else {
+      //TODO: the below gives warning C4244: 'argument': conversion from 'v_int64' to 'TValueType', possible loss of data
```
lganzzzo, line 136: Example of data loss here:

```cpp
oatpp::String json = "256";
auto value = jsonObjectMapper->readFromString<oatpp::UInt8>(json); // UInt8 will be rounded to 0
```

The question is - do we care about such cases?

I think for now - we can just static cast the value to silent the error, and leave the corresponding `TODO info here`

src/oatpp/web/server/HttpProcessor.cpp
```diff
@@ -125,19 +125,19 @@ bool HttpProcessor::processNextRequest(ProcessingResources& resources) {
 
   } catch (oatpp::web::protocol::http::HttpError& error) {
 
-    auto response = resources.components->errorHandler->handleError(error.getInfo().status, error.getMessage(), error.getHeaders());
-    response->send(resources.connection.get(), &resources.headersOutBuffer, nullptr);
+    auto response1 = resources.components->errorHandler->handleError(error.getInfo().status, error.getMessage(), error.getHeaders());
```
lganzzzo, line 128: Maybe, let's just reuse the `response` variable declared before the `try` block.
just to avoid the `..1` vars.

src/oatpp/web/server/api/ApiController.hpp
```diff
@@ -485,6 +485,7 @@ template<>
 struct ApiController::TypeInterpretation <oatpp::Int8> {
   static oatpp::Int8 fromString(const oatpp::String& typeName, const oatpp::String& text, bool& success) {
     (void) typeName;
+    //TODO: the below gives warning C4244: 'argument': conversion from 'v_int32' to 'TValueType', possible loss of data
```
lganzzzo, line 488: This is the same case as in the JSON deserializer.
We could probably check that the value actually falls in the range of `Int8`, but it's out-of-scope for now.

Let's just static cast it to silent the error, and leave the corresponding `TODO`

src/oatpp/core/async/worker/IOEventWorker_common.cpp
srogatch: Is it fine to `#include <Windows.h>` at that point in order to get `INVALID_SOCKET` constant?

src/oatpp/core/data/buffer/FIFOBuffer.cpp
srogatch: Is my fix correct?

src/oatpp/core/data/mapping/type/Any.cpp
srogatch: I'm moving your comment to the code then.

src/oatpp/network/client/SimpleTCPConnectionProvider.cpp
srogatch: Sure

src/oatpp/parser/json/Utils.cpp
srogatch: In order for `i` variable not to conflict with a next loop `for(v_buff_size i = pos; i < result->getSize(); i++) {`. Without the block, there was a warning about `i` variable hiding the definition from the outer scope.

src/oatpp/core/async/worker/IOEventWorker_common.cpp
lganzzzo: Yes, you can include windows.h in `IODefinitions.hpp` file the v_io_handle type is defined there, so it's a good idea to define the corresponding `INVALID_IO_HANDLE` constant in the same place. Then we can use the `INVALID_IO_HANDLE` in the rest of the code.

src/oatpp/core/data/buffer/FIFOBuffer.cpp
lganzzzo: Yep, your fix is correct!

src/oatpp/web/server/HttpProcessor.cpp
srogatch: I thought that may change `shared_ptr` destruction order - if we reuse the variable, the old `shared_ptr` may be destructed inside the catch blocks. Is this fine?

srogatch:
Oops, latest standard draft compiles without `typename` keyword, thus builds on Ubuntu failed. I'm fixing this now.

lganzzzo:
I can see the ubuntu build stuck. It might be the problem with the test(stopping server).
I'll check it once I get to my laptop and merge the PR.

Everything looks great to me at the moment!

lganzzzo:
Great Work!
Merged!

PR #266, opened 2020-06-27
IODefinitions. Move #include <WinSock2.h> back to .cpp file.

lganzzzo:
The `#include <WinSock2.h>` introduces `min` and `max` macros that clush with c++ `numeric_limits::min/max`.
To fix that I'm moving `WinSock2.h` include back to .cpp file. 
And the `INVALID_IO_HANDLE ` constant is now just manually defined as -1 which is rounded to max int32/int64.


bamkrs:
Man, windows is a pain :D

PR #268, opened 2020-06-30
Add "description" argument with a default value to Endpoint::addConsu…

gooichi:
Endpoint.hpp:

```cpp
template<class Wrapper>
void addConsumes(const oatpp::String& contentType, const oatpp::String& description = oatpp::String())
```

Result (Swagger-UI):

Result


lganzzzo:
Nice!
Merged!

gooichi:
Thx! ;-)

Issue #261, opened 2020-06-26
Add a way to disable/change loglevels during runtime


zingmars:
Currently there appear to be two ways to disable logging - one is to #define certain variables during compilation, another one is to subclass the logger class and write your own log function that can be disabled. It would be really useful if there was a way to disable logging using built-in functionality instead, either by setting minimum log level to print or by disabling individual log levels like you can with the compile-time #defines. 

Implementation idea from @oatpp_io_twitter on gitter: https://gitter.im/oatpp-framework/Lobby?at=5ef5d5a9bb149531edd0a4f7

PR #270, opened 2020-07-01
Proposal for #261: Fine grained configurable Loglevel

bamkrs:
We received several wishes to implement a runtime configuration of enabled log-level (the latest one being #261).
In this PR, the individual loglevels are converted to a bitmask rather than an ongoing number. Thus we can enable and disable single log levels without side-affects for the other levels.

However, to keep the `Logger` interface as generic as possible, this is only implemented for the `DefaultLogger`, not the interface. The new function `oatpp::base::Environment::getLogger()` is generic as well and only returns `std::shared_ptr<Logger>`. Therefore, the user has to use a pointer cast to `std::shared_ptr<DefaultLogger>` to be able to call the `disablePriority()` and `enablePriority()` functions. This is a design limitation to keep everything generic and extensible. 

```c++
auto logger = std::static_pointer_cast<oatpp::base::DefaultLogger>(oatpp::base::Environment::getLogger());
logger->disablePriority(oatpp::base::DefaultLogger::PRIORITY_D | oatpp::base::DefaultLogger::PRIORITY_I);
logger->enablePriority(oatpp::base::DefaultLogger::PRIORITY_D | oatpp::base::DefaultLogger::PRIORITY_I);
```

Multiple priorities can be or'd at once.

### DIFF
--- a/src/oatpp/core/base/Environment.cpp
+++ b/src/oatpp/core/base/Environment.cpp
@@ -60,10 +60,6 @@ DefaultLogger::DefaultLogger(const Config& config)
 
 void DefaultLogger::log(v_uint32 priority, const std::string& tag, const std::string& message) {
 
-  if (!(m_config.logMask & priority)) {
-    return;
-  }
-
   bool indent = false;
   auto time = std::chrono::system_clock::now().time_since_epoch();
 
@@ -124,14 +120,20 @@ void DefaultLogger::log(v_uint32 priority, const std::string& tag, const std::st
 
 }
 
-void DefaultLogger::enablePriority(v_uint32 priorities) {
-  m_config.logMask |= priorities;
+void DefaultLogger::enablePriority(v_uint32 priority) {
+  m_config.logMask |= (1 << priority);
 }
 
-void DefaultLogger::disablePriority(v_uint32 priorities) {
-  m_config.logMask &= ~priorities;
+void DefaultLogger::disablePriority(v_uint32 priority) {
+  m_config.logMask &= ~(1 << priority);
 }
 
+bool DefaultLogger::isLogPriorityEnabled(v_uint32 priority) {
+  if (priority > PRIORITY_E) {
+    return true;
+  }
+  return m_config.logMask & (1 << priority);
+}
 
 void Environment::init() {
   init(std::make_shared<DefaultLogger>());
@@ -287,15 +289,34 @@ void Environment::log(v_int32 priority, const std::string& tag, const std::strin
 }
 
 void Environment::logFormatted(v_int32 priority, const std::string& tag, const char* message, ...) {
-  if(message == nullptr) {
-    message = "[null]";
+  // do we have a logger and the priority is enabled?
+  if (m_logger == nullptr || !m_logger->isLogPriorityEnabled(priority)) {
+    return;
   }
-  char buffer[4097];
+  // if we dont need to format anything, just print the message
+  if(message == nullptr) {
+    log(priority, tag, "[null]");
+    return;
+  }
+  // check how big our buffer has to be
   va_list args;
-  va_start (args, message);
-  vsnprintf(buffer, 4096, message, args);
-  log(priority, tag, buffer);
+  va_start(args, message);
+  v_buff_size allocsize = vsnprintf(nullptr, 0, message, args) + 1;
   va_end(args);
+  // alloc the buffer (or the max size)
+  if (allocsize > m_logger->getMaxFormattingBufferSize()) {
+    allocsize = m_logger->getMaxFormattingBufferSize();
+  }
+  char* buffer = new char[allocsize];
+  memset(buffer, 0, allocsize);
+  // actually format
+  va_start(args, message);
+  vsnprintf(buffer, allocsize, message, args);
+  // call (user) providen log function
+  log(priority, tag, buffer);
+  // cleanup
+  va_end(args);
+  delete[] buffer;
 }
 
 void Environment::registerComponent(const std::string& typeName, const std::string& componentName, void* component) {
--- a/src/oatpp/core/base/Environment.cpp
+++ b/src/oatpp/core/base/Environment.cpp
@@ -113,10 +113,15 @@ void DefaultLogger::log(v_uint32 priority, const std::string& tag, const std::st
     indent = true;
   }
 
-  if(indent) {
+  if (indent) {
     std::cout << "|";
   }
-  std::cout << " " << tag << ":" << message << std::endl;
+
+  if (message.empty()) {
+    std::cout << " " << tag << std::endl;
+  } else {
+    std::cout << " " << tag << ":" << message << std::endl;
+  }
 
 }
 
@@ -295,7 +300,7 @@ void Environment::logFormatted(v_int32 priority, const std::string& tag, const c
   }
   // if we dont need to format anything, just print the message
   if(message == nullptr) {
-    log(priority, tag, "[null]");
+    log(priority, tag, std::string());
     return;
   }
   // check how big our buffer has to be
--- a/src/oatpp/core/base/Environment.cpp
+++ b/src/oatpp/core/base/Environment.cpp
@@ -7,6 +7,7 @@
  *
  *
  * Copyright 2018-present, Leonid Stryzhevskyi <lganzzzo@gmail.com>
+ *                         Benedikt-Alexander Mokroß <oatpp@bamkrs.de>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


lganzzzo:
Another thing I was thinking about is the log synchronization...

Currently, we are doing synchronization on the `Logger`. However, it might be a good idea to
guarantee that the `Logger::log(...)` method is thread-safe in order to use fewer resources.

Consider the following example:  
We are logging multiple events from multiple threads and 
we have the following env method called:

```cpp
void Environment::logFormatted(v_int32 priority, const std::string& tag, const char* message, ...) {
    if(message == nullptr) {
      message = "[null]";
    }
    char buffer[4097]; //<--- I don't like this guy it's inconfigurable an it gets allocated for each thread in multithreaded env.
    va_list args;
    va_start (args, message);
    vsnprintf(buffer, 4096, message, args);
    log(priority, tag, buffer); // <--- Call Logger <--- Call Logger <--- Call Logger
    va_end(args);
}
```

- Then if we have synchronization on the `Logger::log()` level we'll allocate the formatting `buffer` for each threat, consuming unnecessary resources.

- But if we'll synchronize on the `Environment::logFormatted(...)` level - then we have only one buffer allocated for a single thread that acquired the mutex.

Something like this:

```cpp
void Environment::logFormatted(v_int32 priority, const std::string& tag, const char* message, ...) {
    if(message == nullptr) {
      message = "[null]";
    }
    {
      std::lock_guard<std::mutex> lock(m_logMutex);
      char buffer[m_logger->logBufferSize() + 1]; //<--- Now we can configure it because we know exactly how much memory is allocated for logging!
      va_list args;
      va_start (args, message);
      vsnprintf(buffer, m_logger->logBufferSize(), message, args);
      log(priority, tag, buffer); // <--- Call Logger <--- Call Logger <--- Call Logger
      va_end(args);
    }
}
```

bamkrs:
> ## Log Priorities
> 
> It's better to keep the log `priority` as continuously increasing numbers.
> Having log priorities as orders of 2 may confuse the user that the actual log `priority` value is
> a mask and that it's possible to mix multiple log levels in it.

My only motivation for this was

* Bitshifting done at compile time (ok, not that bitshifting is a heavy instruction anyways :D)
* Being able to enable multiple levels with one call

However, I consider both as cosmetics... So no problem here :)

> ## Logger Interface
> 
> I propose to extend the logger interface with the following method:
> 
> ```c++
> virtual bool isLogLevelEnabled(v_uint32 priority) const { 
>   return true; // default implementation returns true for all log-levels
> }
> ```

I did not want to touch the very generic interface. When we are touching it we can also add the enable/disable call anyways.

> 
> This will give us the possibility to disable logs before the string formatting here - https://github.com/oatpp/oatpp/blob/master/src/oatpp/core/base/Environment.cpp#L272
> 
> Thus the user will be able to disable logs in his own implementation before the time consuming `vsnprintf` is called and before the log buffer is allocated.

That is indeed a good consideration. :+1: 

> ## Priority --> Level
> 
> Maybe it's better to rename the log `priority` to log `level`?

I think we have to use multiple names for different meanings:

* priority: The *priority* of the current message
* level: The current level a message-priority is required to be printed
* mask: The "filter" which is used to determine if a message-priority is printed or not

----

> Another thing I was thinking about is the log synchronization...
> 
> Currently, we are doing synchronization on the `Logger`. However, it might be a good idea to
> guarantee that the `Logger::log(...)` method is thread-safe in order to use fewer resources.

I'm really not a fan of synchronizing the Logger. There might be a case where some users have threads with heavy logging output. This thread would most probably block a lot of other threads that are logging, too. 

However, there are some other ways we could address the large allocation for the logging memory:

```c++
void Environment::logFormatted(v_int32 priority, const std::string& tag, const char* message, ...) {
    if (!m_logger || !m_logger->isLogLevelEnabled(priority)) {
        return;
    }
    if(message == nullptr) {
      log(priority, tag, "[null]");
    }
    va_list args;
    va_start (args, message);
    v_io_size allocsize = vsnprintf(nullptr, 0, message, args);
    va_end(args); // I don't know if we actually need this here
    if (allocsize > MAX_BUFFER_ALLOC_SIZE) {
        allocsize = MAX_BUFFER_ALLOC_SIZE;
    }
    std::vector<char> buffer(allocsize + 1, 0);
    va_start (args, message);
    vsnprintf(buffer.data(), allocsize, message, args);
    log(priority, tag, buffer.data()); // <--- Call Logger <--- Call Logger <--- Call Logger
    va_end(args);
}
```

We trade memory-performance with cpu-performance. First we check how much space we have to allocate. Then we check if it exceeds a maximum allocation size (could be a `#define` or a call to any function)

lganzzzo:
> I'm really not a fan of synchronizing the Logger. There might be a case where some users have threads with heavy logging output. This thread would most probably block a lot of other threads that are logging, too.

I agree, synchronizing the logger on the framework level may be a bad idea.
While we still have to sync it when writing to `std::cout`, the end-user may want to write it in whatever storage or log-collecting software. So you are right, it's better to leave it the way it is now.

> We trade memory-performance with cpu-performance. First we check how much space we have to allocate. Then we check if it exceeds a maximum allocation size (could be a #define or a call to any function). Only then we are allocating the memory (I used a std::vector for this so we don't have to deal with new/delete[]) and calling the user defined logging function.

I think it's expected that logging downgrades the performance. - I'm ok to trade performance in favor of memory here.

> (could be a #define or a call to any function)

let's put it in the Logger interface with the default implementation provided

```cpp
virtual v_buff_size getMaxFormattingBufferSize() {
  return 4096;
}
```

bamkrs:
> > I'm really not a fan of synchronizing the Logger. There might be a case where some users have threads with heavy logging output. This thread would most probably block a lot of other threads that are logging, too.
> 
> I agree, synchronizing the logger on the framework level may be a bad idea.
> While we still have to sync it when writing to `std::cout`, the end-user may want to write it in whatever storage or log-collecting software. So you are right, it's better to leave it the way it is now.
> 
Yes, but I think the impact on syncing while printing is much smaller then syncing the entire process.
We could even synchronize an even smaller part of the printing if `DefaultLogger::Log` would first write everything to a `std::stringstream` and then just forward the stream to `std::out`. However, we would need to `#include <stringstream>` in the environment header. A header that is included nearly in everything that the user will implement using Oat++. I don't like this. We would pull an unnecessary amount of code in each user module. 

This is the same reason I did not use `std::vector<char>` in `logFormatted` here: https://github.com/oatpp/oatpp/pull/270/commits/ad69f19757b1406efa8847250bae037a4a6f3e0e#diff-152e3d0dd4933f459410ea0ee635a643R310
I wasn't comfortable with this. Even if its an >=C++11 Framework :)

Anyhow, I implemented your feedback in https://github.com/oatpp/oatpp/pull/270/commits/ad69f19757b1406efa8847250bae037a4a6f3e0e and improved the logging behaviour in https://github.com/oatpp/oatpp/pull/270/commits/4e218c5b6b6dae6b1dd9eab04171abf9bfd816f4 in case some users don't want to use tags. However, I could roll back this commit if you don't like it.

src/oatpp/core/base/Environment.cpp
```diff
@@ -270,15 +294,34 @@ void Environment::log(v_int32 priority, const std::string& tag, const std::strin
...
-  va_start (args, message);
-  vsnprintf(buffer, 4096, message, args);
+  va_start(args, message);
+  v_buff_size allocsize = vsnprintf(nullptr, 0, message, args) + 1;
+  va_end(args);
+  // alloc the buffer (or the max size)
+  if (allocsize > m_logger->getMaxFormattingBufferSize()) {
+    allocsize = m_logger->getMaxFormattingBufferSize();
+  }
+  char* buffer = new char[allocsize];
```
lganzzzo, line 315: Let's use the `unique_ptr` here, as the user-provided `log()` may potentially throw.

lganzzzo:
Looks great to me.
The only comment - is to use the `unique_ptr` for dynamic buffer.

bamkrs:
> The only comment - is to use the `unique_ptr` for dynamic buffer.

I want to avoid using `std::unique_ptr` since its heavily discouraged to do so. And since std::vector is out of scope, too. We *could* use std::string:

```c++
std::string buffer;
buffer.resize(allocsize, 0);
memset((char*)buffer.data(), 0, allocsize);
// actually format
va_start(args, message);
vsnprintf((char*)buffer.data(), allocsize, message, args);
``` 

**But** this isn't nice too. You see `buffer.data()` returns a `const char*` and its just casted to a `char*`... I don't know how save this is (probably not that much...)

Your call, `std::unique_ptr` or hacky `std::string`?

lganzzzo:
I would opt for `std::unique_ptr`.
To be honest, I don't see any cons of using it here.
We have already `memory` include in the `Environment.hpp` and we are using smart pointers pretty extensively through our code.

lganzzzo:
Great Work!
Merged!

PR #273, opened 2020-07-12
Fix warning that m_currentResult is not initialized in the constructor

hyperxor:


lganzzzo:
Nice!
Thanks for contributing!

hyperxor:
@lganzzzo you are welcome!

PR #274, opened 2020-07-13
fix cannot serialize 'false' value

NickPak:


bamkrs:
Nice catch! However, we can not merge it this way. 
Just removing the check will cause horrible problems down the line.

You could rewrite the if statement to something like `if (parameter != nullptr)` or `if (parameter.get() != nullptr)`.
This would fix the issue and keep the safeguard in place.

NickPak:
You are right. I've rewritten it. In order for the `toString` method never to return a `nullptr`, I returned a default value of 'false', even though this code would almost never be executed.
```
    if(parameter.get() != nullptr) {
        return utils::conversion::boolToStr(*parameter);
    }
    return oatpp::String("false", 5, false);
```

lganzzzo:
Hey guys,

Here is the table of the `oatpp::Boolean` conversions:

```cpp
{
  oatpp::Boolean b; // nullptr;
  OATPP_ASSERT(!b); // resolves to false; Safe conversion.
  OATPP_ASSERT(b == nullptr); // it IS equal to nullptr
}

{
  oatpp::Boolean b = false; // not null, value=false;
  OATPP_ASSERT(!b); // resolves to false;
  OATPP_ASSERT(b != nullptr); // b is NOT equal to nullptr;
}

{
  oatpp::Boolean b = true;
  OATPP_ASSERT(b); // resolves to true;
  OATPP_ASSERT(b != nullptr); // b is NOT nullptr;
}
```

## The Original Issue

> When I declared a Boolean argument using the QUERY macro and tried to pass the false value, it caused a crash

As it was correctly pointed by @NickPak , the issue is [here](https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/client/ApiClient.hpp#L392) in the `ApiClient::TypeInterpretation`.

The `if` statement is incorrect for boolean here and a valid `false` value leads to the `nullptr` being returned.

## The Fix

The correct fix here I think would be:

```cpp
template<>
struct ApiClient::TypeInterpretation<oatpp::Boolean> {
  static oatpp::String toString(const oatpp::String &typeName, const oatpp::Boolean &parameter) {
    (void) typeName;
    if (parameter != nullptr) { //<--- Check for nullptr.
      return utils::conversion::boolToStr(*parameter);
    }
    return nullptr; //<--- still return the nullptr if nullptr was passed
  }
};
```

src/oatpp/core/data/mapping/type/Primitive.hpp
```diff
@@ -298,7 +298,7 @@ class Boolean : public type::ObjectWrapper<bool, __class::Boolean>  {
     return !operator == (other);
   }
 
-  inline operator bool() const {
+  explicit inline operator bool() const override {
```
lganzzzo, line 301: Please do not make it `explicit`.
See the `oatpp::Boolean` conversion table in the [comment](https://github.com/oatpp/oatpp/pull/274#issuecomment-658173567)

src/oatpp/core/data/mapping/type/Type.hpp
```diff
@@ -200,7 +200,7 @@ class ObjectWrapper {
     return m_ptr.get() != other.m_ptr.get();
   }
   
-  explicit inline operator bool() const {
+  explicit virtual inline operator bool() const {
```
lganzzzo, line 203: Please remove `virtual`. 
This functionality is not in the context of this issue.
Also, we want to avoid virtual methods in the `ObjectWrapper`.
We want to keep `sizeof` ObjectWrapper as small as possible.

src/oatpp/web/client/ApiClient.hpp
```diff
@@ -389,10 +389,10 @@ template<>
 struct ApiClient::TypeInterpretation<oatpp::Boolean> {
   static oatpp::String toString(const oatpp::String &typeName, const oatpp::Boolean &parameter) {
     (void) typeName;
-    if (parameter) {
-      return utils::conversion::boolToStr(*parameter);
+    if(parameter.get() != nullptr) {
+        return utils::conversion::boolToStr(*parameter);
     }
-    return nullptr;
+    return oatpp::String("false", 5, false);
```
lganzzzo, line 395: I believe it's better to return `nullptr` if `nullptr` was passed.
I agree that the silent crash is not good, but as for this exact pice of functionality, returning `nullptr`
is expected behavior.

NickPak:
You are right. I've rewritten it.

lganzzzo:
Great work! 
Merged!

PR #276, opened 2020-07-15
Initial implementation of defaults for swagger

natanael:
Following the issue https://github.com/oatpp/oatpp-swagger/issues/23
I don't think this serves as a complete solution for the defaults in swagger, as it should match the type of the field and I implemented it as a string in all cases. The word `default` is quite tricky as well so I went with `default_value`, if anything it makes it more explicit.
I'm open to all suggestions.

### DIFF
--- a/src/oatpp/core/data/mapping/type/Type.hpp
+++ b/src/oatpp/core/data/mapping/type/Type.hpp
@@ -280,7 +280,7 @@ public:
        * Description.
        */
       std::string description = "";
-      std::string default_value = "";
+      std::string defaultValue = "";
     };
 
   private:
--- a/src/oatpp/core/data/mapping/type/Type.hpp
+++ b/src/oatpp/core/data/mapping/type/Type.hpp
@@ -280,6 +280,7 @@ public:
        * Description.
        */
       std::string description = "";
+      std::string default_value = "";
     };
 
   private:
--- a/test/oatpp/core/data/mapping/type/ObjectTest.cpp
+++ b/test/oatpp/core/data/mapping/type/ObjectTest.cpp
@@ -47,6 +47,7 @@ class DtoA : public oatpp::DTO {
 
   DTO_FIELD_INFO(id) {
     info->description = "identifier";
+    info->default_value = "Some default id";
   }
   DTO_FIELD(String, id) = "Some default id";
 
@@ -66,6 +67,7 @@ class DtoB : public DtoA {
 
   DTO_FIELD_INFO(a) {
     info->description = "some field with a qualified name";
+    info->default_value = "default-value";
   }
   DTO_FIELD(String, a, "field-a") = "default-value";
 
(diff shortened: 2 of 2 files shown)


src/oatpp/core/data/mapping/type/Type.hpp
```diff
@@ -280,6 +280,7 @@ class Type {
        * Description.
        */
       std::string description = "";
+      std::string default_value = "";
```
lganzzzo, line 283: Please rename to `defaultValue`
Oat++ has a "non-standard" "legacy" code-style.

lganzzzo:
> I don't think this serves as a complete solution for the defaults in swagger, as it should match the type of the field and I implemented it as a string in all cases.

Yes, it's a tricky one.
We could possibly use `oatpp::Any` here - but the problem is that `OATPP_FIELD_INFO` - is set during the static initialization.
And storing oatpp mapping-enabled objects in as static objects will cause tests to fail due to object counting (mem leak detection)

lganzzzo:
I have a new idea.

It should be possible to do the following:

```cpp
class MyDto : oatpp::DTO {

  ...

  DTO_FIELD_INFO(someValue) {
    info->description = "some value";
    info->hasDefaultValue = true; //<--- always true by default
  }
  DTO_FIELD(oatpp::String, someValue) = "default-value"; //<--- let's detect the default value from here!

};
```

## How it gonna work

Each DTO object has a creator function defined by it's `__class`.
In `oatpp-swagger` during the documentation process of DTO, we can call DTO creator function and obtain an object in
its default state.
Thus, by scanning, its fields, and values - we know exactly what the default field-values are.

Actually swagger module already does creation and scanning - so all we need - is to document the value.

[Here](https://github.com/oatpp/oatpp-swagger/blob/master/src/oatpp-swagger/oas3/Generator.cpp#L128) we should also instantiate an object by calling the creator:

```cpp
    auto instance = type->creator(); //<--- Create instance of the object.
    auto properties = type->propertiesGetter();

    for(auto* p : properties->getList()) {
      
      result->properties[p->name] = generateSchemaForType(p->type, 
                                                          true, 
                                                          usedTypes, 
                                                          p, 
                                                          p->get(instance)); //<--- Pass the default value of the property.
    }
```

Thus, having a default value as an oatpp mapping-enabled type - we can properly document the default value for any given type.

src/oatpp/core/data/mapping/type/Type.hpp
natanael: My bad! True, everything is camel cased. I just fixed it.

natanael:
Continuing at https://github.com/oatpp/oatpp-swagger/pull/24

src/oatpp/core/data/mapping/type/Type.hpp
```diff
@@ -280,6 +280,7 @@ class Type {
        * Description.
        */
       std::string description = "";
+      std::string defaultValue = "";
```
lganzzzo, line 283: Let's change it to something like `bool hasDefaultValue = true`

lganzzzo:
I'm closing this PR.
- Swagger docs for default values are implemented in https://github.com/oatpp/oatpp-swagger/pull/24
- The last proposition of mine is minor and I doubt if it's really needed.

PR #282, opened 2020-07-28
Adds pattern to the info, to be used in swagger

natanael:
Just the first step to have something like the following working:
```cpp
DTO_FIELD_INFO(start_date) {
    info->description = "Your start date";
    info->pattern = "^\d{4}-\d{2}-\d{2}$";
}
DTO_FIELD(String, start_date);
```
So in `oatpp-swagger` we can just pass it through and make the OAS like:
```json
"start-date": {
    "type": "string",
    "pattern": "^\d{4}-\d{2}-\d{2}$" // <- some swagger clients may now validate the string date with this
},
```

### DIFF
--- a/src/oatpp/core/data/mapping/type/Type.hpp
+++ b/src/oatpp/core/data/mapping/type/Type.hpp
@@ -280,6 +280,7 @@ public:
        * Description.
        */
       std::string description = "";
+      std::string pattern = "";
     };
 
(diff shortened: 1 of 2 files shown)


lganzzzo:
Great idea,
Merged!

Please also update `oatpp-swagger` - to use this value!

PR #287, opened 2020-08-01
Draft for an optional non-blocking call to `Server::run`

bamkrs:
This hides the needed threading to make the server non-blocking for
inexperienced users or those who do not want to run their own threading
concept.

### DIFF
--- a/src/oatpp/network/server/Server.cpp
+++ b/src/oatpp/network/server/Server.cpp
@@ -73,8 +73,12 @@ void Server::run(bool blocking) {
     setStatus(STATUS_CREATED, STATUS_STARTING);
     mainLoop(this);
   } else {
-    if (getStatus() == STATUS_RUNNING)
-      return;
+    switch (getStatus()) {
+      case STATUS_STARTING:
+      case STATUS_RUNNING:
+        // Maybe throw an exception here instead of just silently return?
+        return;
+    }
     m_blocking = blocking;
     setStatus(STATUS_CREATED, STATUS_STARTING);
     m_thread = std::thread(mainLoop, this);
--- a/src/oatpp/network/server/Server.cpp
+++ b/src/oatpp/network/server/Server.cpp
@@ -28,95 +28,95 @@
 #include <chrono>
 
 namespace oatpp { namespace network { namespace server {
-  
+
 const v_int32 Server::STATUS_CREATED = 0;
 const v_int32 Server::STATUS_STARTING = 1;
 const v_int32 Server::STATUS_RUNNING = 2;
 const v_int32 Server::STATUS_STOPPING = 3;
 const v_int32 Server::STATUS_DONE = 4;
 
-Server::Server(const std::shared_ptr<ServerConnectionProvider>& connectionProvider,
-               const std::shared_ptr<ConnectionHandler>& connectionHandler)
-  : m_status(STATUS_CREATED)
-  , m_connectionProvider(connectionProvider)
-  , m_connectionHandler(connectionHandler)
-  , m_blocking(false)
-{}
+Server::Server(const std::shared_ptr<ServerConnectionProvider> &connectionProvider,
+               const std::shared_ptr<ConnectionHandler> &connectionHandler)
+    : m_status(STATUS_CREATED)
+    , m_connectionProvider(connectionProvider)
+    , m_connectionHandler(connectionHandler)
+    , m_threaded(false) {}
 
-void Server::mainLoop(Server *instance){
-  
+void Server::mainLoop(Server *instance) {
+
   instance->setStatus(STATUS_STARTING, STATUS_RUNNING);
 
   std::shared_ptr<const std::unordered_map<oatpp::String, oatpp::String>> params;
 
-  while(instance->getStatus() == STATUS_RUNNING) {
-    
+  while (instance->getStatus() == STATUS_RUNNING) {
+
     auto connection = instance->m_connectionProvider->getConnection();
 
     if (connection) {
-      if(instance->getStatus() == STATUS_RUNNING){
+      if (instance->getStatus() == STATUS_RUNNING) {
         instance->m_connectionHandler->handleConnection(connection, params /* null params */);
       } else {
         OATPP_LOGD("Server", "Already stopped. Closing connection...");
       }
     }
-    
+
   }
 
   instance->setStatus(STATUS_DONE);
 
 }
 
-void Server::run(bool blocking) {
-  if (blocking) {
-    m_blocking = blocking;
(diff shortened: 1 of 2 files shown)


src/oatpp/network/server/Server.cpp
```diff
@@ -30,47 +30,62 @@
...
-  while(getStatus() == STATUS_RUNNING) {
+  while(instance->getStatus() == STATUS_RUNNING) {
     
-    auto connection = m_connectionProvider->getConnection();
+    auto connection = instance->m_connectionProvider->getConnection();
 
     if (connection) {
-      if(getStatus() == STATUS_RUNNING){
-        m_connectionHandler->handleConnection(connection, params /* null params */);
+      if(instance->getStatus() == STATUS_RUNNING){
```
bamkrs, line 57: This should also check for `STATUS_STARTING`.
And maybe we could throw an exception or log a warning here instead of silently return.

src/oatpp/network/server/Server.cpp
```diff
@@ -30,47 +30,66 @@
...
+void Server::run(bool blocking) {
+  if (blocking) {
+    m_blocking = blocking;
+    setStatus(STATUS_CREATED, STATUS_STARTING);
+    mainLoop(this);
+  } else {
+    switch (getStatus()) {
+      case STATUS_STARTING:
+      case STATUS_RUNNING:
+        // Maybe throw an exception here instead of just silently return?
```
bamkrs, line 79: Maybe throw an exception here instead of just silently return?

src/oatpp/network/server/Server.cpp
```diff
@@ -30,47 +30,66 @@
...
+        return;
```
bamkrs, line 80: Maybe throw an exception here instead of just silently return?

lganzzzo: I agree - throw is much better here

src/oatpp/network/server/Server.cpp
```diff
@@ -30,47 +30,66 @@
...
-  setStatus(STATUS_DONE);
-  
+
+  instance->setStatus(STATUS_DONE);
+
 }
 
-void Server::run(){
-  mainLoop();
+void Server::run(bool blocking) {
```
lganzzzo, line 70: The name "blocking"/"non-blocking" is confusing to me in this context.
I think it's better to describe exactly what it's going to do:

```cpp
Server::run(bool createNewThread);
```

src/oatpp/network/server/Server.cpp
```diff
@@ -30,47 +30,66 @@
...
+        return;
+    }
+    m_blocking = blocking;
+    setStatus(STATUS_CREATED, STATUS_STARTING);
+    m_thread = std::thread(mainLoop, this);
+  }
 }
   
 void Server::stop(){
   setStatus(STATUS_STOPPING);
```
lganzzzo, line 89: What if we have stop method like this:

```cpp
void Server::stop(bool wait = true);
```

if wait == true then wait for the server to stop.

src/oatpp/network/server/Server.cpp
```diff
@@ -86,5 +105,18 @@ v_int32 Server::getStatus(){
...
 
+Server::~Server() {
+  switch (getStatus()) {
+    case STATUS_STARTING:
+    case STATUS_RUNNING:
+      stop();
+      break;
+    default:
+      // Just check if the thread is still alive and wait for it to exit
+      if (!m_blocking && m_thread.joinable()) {
```
lganzzzo, line 116: Let's keep all `stop()` relates checks in the stop method.
Let's have the destructor to be as follows:

```cpp
Server::~Server() {
  stop(); // wait here if server was not stopped
}
```

src/oatpp/network/server/Server.hpp
```diff
@@ -103,8 +114,9 @@ class Server : public base::Countable {
   /**
    * Call &id:oatpp::network::ConnectionProvider::getConnection; in the loop and passes obtained Connection
    * to &id:oatpp::network::server::ConnectionHandler;.
+   * @param blocking - Start the server blocking (thread of callee) or non-blocking (own thread)
    */
-  void run();
+  void run(bool blocking = true);
```
lganzzzo, line 119: The name "blocking"/"non-blocking" is confusing to me in this context.
I think it's better to describe exactly what it's going to do:

```cpp
Server::run(bool createNewThread);
```

src/oatpp/network/server/Server.cpp
bamkrs: wouldn't be `createAsThread` or `createInNewThread` a better fit?

src/oatpp/network/server/Server.cpp
bamkrs: If we do not wait for the thread to stop, we may run in the problem where the server-object gets deallocated while the thread (that has a reference to the server-object) still does something and tries to access it. This would cause havoc :100:

src/oatpp/network/server/Server.hpp
bamkrs: See cpp

src/oatpp/network/server/Server.cpp
lganzzzo: I agree - createInNewThread a better fit

src/oatpp/network/server/Server.cpp
lganzzzo: Agree. Let's then always wait for the thread to finish.

Also, I was thinking of `condition_variable` here with detached thread instead of:

```cpp
  if (!m_blocking && m_thread.joinable()) {
      m_thread.join();
  }
```

To make `stop()` to work consistently event if called from multiple threads.
But I would also agree that it's not a big deal at the moment.

lganzzzo:
> @bamkrs bamkrs deleted a comment from lganzzzo 24 minutes ago

Something was deleted here 😃

src/oatpp/network/server/Server.cpp
bamkrs: We could guard it with a mutex. Both start and stop. Using `std::muted.try_lock()` to check if it is already stopping etc

lganzzzo: Agree

bamkrs:
> > @bamkrs bamkrs deleted a comment from lganzzzo 24 minutes ago
> 
> Something was deleted here smiley

yeah, it was a duplicated review entry. dunno what went wrong there..?

bamkrs:
@lganzzzo implemented the review and maybe deviated a bit here and there

lganzzzo:
Looks great!

PR #297, opened 2020-09-08
add the host parameter

NickPak:

PR #298, opened 2020-09-09
Pr #297 review fixes

lganzzzo:


lganzzzo:
Hey @NickPak ,

This branch contains some fixes/improvements following your original PR(#297)

NickPak:
@lganzzzo I'm so sorry. I just saw this error. By the time I was ready to submit the fixes, you had already fixed them. Yes, you are right, because it is bind to an IPv6 address by default, so the client can no longer use 127.0.0.1

PR #299, opened 2020-09-09
Remove directory versioning during install. GCC not always could find …

nicraMarcin:
…headers in oatpp/oatpp-{VERSION} directory


lganzzzo:
Hey @nicraMarcin ,

Please provide more detailed reasoning for removing version suffixes.

Originally directory version suffixes were intentionally added to keep module installations consistent during upgrades,
and to make it easier to keep track of installed modules and their versions.

If there is no version suffix you won't have old files deleted during the upgrade. You can't have multiple oatpp versions installed on the same machine (useful for testing). It's hard to tell which modules of which versions you have installed by just looking into installation directories.
It makes the process of upgrading oatpp installation less clear and more error-prone.

Regards,
Leonid

bamkrs:
This looks more like a problem of your compiler setup then an general problem. 
I work with oatpp in multiple versions on multiple architectures and never had trouble with the headers. I would vote to reject the PR.

nicraMarcin:
@bamkrs so now you can tell that I have this issue on two different computers. I try on other my linux machine with mint and I have the same problem in CLion.

@lganzzzo I agree that versioning is good for testing and this is the reason of versioning. Bit if anybody wants different version of library needs to do additional stufs in his project, like writting more lines in cmake to provide strict libraries.

Please try with ubuntu/debian and clion install oatpp-starter.

bamkrs:
@nicraMarcin That is exactly my Setup. I'm working with Oat++ over a year or something, started with Debian 9, later moved to 10 and now I'm running Debian 11. I'm using CLion since 2016 iirc and never had any troubles. The same thing goes for my MacOS setup. No problems since I've begun working on/with Oat++.
We don't receive any bug reports regarding the versioning here or in gitter.io either. The only few we receive are incorrectly set up paths 99,9% of the time (unintentionally installed to wrong path prefix etc).
So it is save to assume, that there must be something you are missing while setting up your system. No offence or anything, its just our experience :)

nicraMarcin:
This is just my suggestion and if the main tutor deems it unnecessary, no problem.

bamkrs:
You could rewrite the PR so that the header-versioning can be disabled by an cmake-option. Wouldn't this be the perfect compromise? Versioning on by default, but optionally disabled.

lganzzzo:
> You could rewrite the PR so that the header-versioning can be disabled by an cmake-option. Wouldn't this be the perfect compromise? Versioning on by default, but optionally disabled.

Not sure if we want such an option. I would prefer to keep the number of possible installation-dir variants as small as possible - exactly `1`.

lganzzzo:
I'm closing this PR.
@nicraMarcin , @bamkrs please feel free to continue the discussion here in case there are other propositions.

PR #302, opened 2020-09-15
Fix for long lasting issue #203 (Closes #203)

### LINKED ISSUE #203: ToDo: Empty path (`""`) in Endpoint causes SEGFAULT in `Router`
In this scenario, we have a controller that does not have any path-prefix:
```c++
MyController(OATPP_COMPONENT(std::shared_ptr<ObjectMapper>, objectMapper))
  : oatpp::web::server::api::ApiController(objectMapper, nullptr){}
```
----

We create an endpoint that should be the index-endpoint like:
```c++
ENDPOINT("GET", "/", index);
```

This works perfectly fine. However, when we omit the `/` in the path and just have an empty string like
```c++
ENDPOINT("GET", "", index);
```
the application SEGFAULTs  when the router tries to access the paths since the `mapping` part of the routing-pair becomes `nullptr` when `toString()` is called.
I.e. in `Router.hpp` when printing the mappings. `pair.first->toString()` returns `nullptr` instead of `""` because it seems that an empty string is converted to a `nullptr` instead of an empty `oatpp::String`.
```c++
for(auto& pair : m_endpointsByPattern) {
      auto mapping = pair.first->toString();
      OATPP_LOGD("Router", "url '%s' -> mapped", (const char*) mapping->getData());
}
```

Since Oat++ is designed in such a way that `ENDPOINT("GET", "/test", test)` is *"the same"* as `ENDPOINT("GET", "test", test)`, one should also be able to use `ENDPOINT("GET", "", index)` for consistency.

----
So it is a bug because it segfaults the application as well as an enhancement since its yet another convenience feature.

bamkrs:
This is just a small workaround for issue #203. Doing a small check here does not introduce any "online"-costs since its just excecuted at startup.
It prevents us from unnecessary checks in the functions that are run every time a new request comes in.

### DIFF
--- a/src/oatpp/codegen/api_controller/base_define.hpp
+++ b/src/oatpp/codegen/api_controller/base_define.hpp
@@ -290,6 +290,9 @@ EndpointInfoBuilder Z__CREATE_ENDPOINT_INFO_##NAME = [this](){ \
   info->name = #NAME; \
   info->path = ((m_routerPrefix != nullptr) ? m_routerPrefix + PATH : PATH); \
   info->method = METHOD; \
+  if (info->path == "") { \
+    info->path = "/"; \
+  } \
   return info; \
 }; \
 \


lganzzzo (approved):
Looks good to me

PR #365, opened 2020-12-29
WIP/Discussion: Let server check an (optional) stopping condition by itself

bamkrs:
Instead of an external invocation of `stop()` to stop a running server, one could pass an lambda that returns true when the server should continue to run or false if it should stop.
Ideally this function should return as fast as possible and only do minimal amount of work. Like an interrupt function in µC.
This would ease the implementation of simple stopping mechanism that do not require external threads or other management.
For more complex functions, the minimal `bool()` function can be bound with `std::bind` to pass parameters etc.

In src/oatpp-test/web/ClientServerTestRunner.hpp a quick and dirty example is shown how a lambda is passed to `run()` which does nothing but return an bool that is modified somewhere other. This part isn't meant to be merged just as a dirty example.

Consider the following pseudo-code:

```c++
void RunApp() {
/* [...] */
  SomeLogicClass slc(); // does some logic, maybe reads some other sockets or whatever in its own thread etc

  std::function<bool()> condition = [&slc]() {
    return slc.HasError() == false;
  };

  /* Create server which takes provided TCP connections and passes them to HTTP connection handler */
  oatpp::network::Server server(connectionProvider, connectionHandler);
  
  server.run(false /* run in this thread */, condition);
}
```

In this example, the condition-lambda checks if an other part of our application logic has an error. If not, the lambda returns true and the server continues to run. If an error occurs, the lambda returns false and the server stops.

There are other ways to archive this behaviour. One could pass an `shared_ptr` to the `SomeLogicClass` which then calls `server.stop()` in its error handler. However, this may not be possible because the `SomeLogicClass` is an blackbox-library. 
Another option would be to have an entire thread just checking the error-status of `SomeLogicClass` and then calls `server.stop()` which might be to big of an overhead.

### DIFF
--- a/src/oatpp-test/web/ClientServerTestRunner.hpp
+++ b/src/oatpp-test/web/ClientServerTestRunner.hpp
@@ -89,21 +89,25 @@ public:
     std::atomic<bool> running(true);
     std::mutex timeoutMutex;
     std::condition_variable timeoutCondition;
+    bool runConditionForLambda = true;
 
     m_server = std::make_shared<oatpp::network::Server>(m_connectionProvider, m_connectionHandler);
     OATPP_LOGD("\033[1;34mClientServerTestRunner\033[0m", "\033[1;34mRunning server on port %s. Timeout %lld(micro)\033[0m",
                m_connectionProvider->getProperty("port").toString()->c_str(),
                timeout.count());
 
-    std::thread serverThread([this]{
-      m_server->run();
-    });
+    std::function<bool()> condition = [runConditionForLambda](){
+        return runConditionForLambda;
+    };
 
-    std::thread clientThread([this, &lambda]{
+    m_server->run(true, condition);
+
+    std::thread clientThread([&runConditionForLambda, this, &lambda]{
 
       lambda();
 
-      m_server->stop();
+//      m_server->stop();
+      runConditionForLambda = false;
       m_connectionHandler->stop();
       m_connectionProvider->stop();
 
@@ -120,7 +124,6 @@ public:
 
     });
 
-    serverThread.join();
     clientThread.join();
 
     auto elapsed = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - startTime);
--- a/src/oatpp-test/web/ClientServerTestRunner.hpp
+++ b/src/oatpp-test/web/ClientServerTestRunner.hpp
@@ -96,7 +96,7 @@ public:
                m_connectionProvider->getProperty("port").toString()->c_str(),
                timeout.count());
 
-    std::function<bool()> condition = [runConditionForLambda](){
+    std::function<bool()> condition = [&runConditionForLambda](){
         return runConditionForLambda;
     };
 
--- a/src/oatpp/network/Server.cpp
+++ b/src/oatpp/network/Server.cpp
@@ -26,6 +26,7 @@
 
 #include <thread>
 #include <chrono>
+#include <utility>
 
 namespace oatpp { namespace network {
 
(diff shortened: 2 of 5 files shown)


src/oatpp/network/Server.cpp
```diff
@@ -45,28 +46,48 @@ Server::Server(const std::shared_ptr<ServerConnectionProvider> &connectionProvid
 void Server::mainLoop(Server *instance) {
```
lganzzzo, line 46: It might be better to split this into 
- loop with condition
- loop without condition

src/oatpp/network/Server.cpp
```diff
@@ -45,28 +46,48 @@ Server::Server(const std::shared_ptr<ServerConnectionProvider> &connectionProvid
...
     }
-
   }
 
   instance->setStatus(STATUS_DONE);
 
 }
 
-void Server::run(bool startAsNewThread) {
+void Server::run(bool startAsNewThread, std::function<bool()> conditional) {
```
lganzzzo, line 90: I would prefer to get rid of a `startAsNewThread` - it's too confusing to me.
Let `Server::run()` always execute on a current thread.



```cpp
void Server::run(std::function<bool()> stopCondition = nullptr);
```

test/oatpp/web/PipelineTest.cpp
```diff
@@ -176,7 +176,7 @@ void PipelineTest::onRun() {
     pipeOutThread.join();
     pipeInThread.join();
 
-  }, std::chrono::minutes(10));
+  }, std::chrono::minutes(1));
```
lganzzzo, line 179: Let's keep this as it previously was. It's not a part of this PR.

bamkrs: This was just part of debugging the tests. The pipeline-tests always time out on my machine (Debian 11) and I decreased it out of curiosity. Has no relation to the PR and will be squashed away. (If the PR will be merged, I would recommend an squash-merge anyway).

src/oatpp/network/Server.cpp
bamkrs: Its a feature we introduced in 1.1 iirc since we received a ton of issues because

> `server.run()` never returns

We have not received any complaints since then. My vote would be for keeping it. Maybe we should do some kind of poll in gitter/github?

src/oatpp/network/Server.cpp
bamkrs: Ha, its funny because I originally just did that and scraped it later on to "not bloat things up".  When the loop logic grew a bit I did not split it up again. Just wanted to hear what you think about this new functionality before optimising things.

src/oatpp-test/web/ClientServerTestRunner.hpp
```diff
@@ -89,21 +89,25 @@ class ClientServerTestRunner {
...
 
     m_server = std::make_shared<oatpp::network::Server>(m_connectionProvider, m_connectionHandler);
     OATPP_LOGD("\033[1;34mClientServerTestRunner\033[0m", "\033[1;34mRunning server on port %s. Timeout %lld(micro)\033[0m",
                m_connectionProvider->getProperty("port").toString()->c_str(),
                timeout.count());
 
-    std::thread serverThread([this]{
-      m_server->run();
-    });
+    std::function<bool()> condition = [runConditionForLambda](){
```
bamkrs, line 99: Not meant to be merged in this PR, this is just an dirty and not well engineered example.

src/oatpp/network/Server.cpp
lganzzzo: Yes, I know. I treated that mostly as a workaround. 
To my mind, it doesn't give any benefits. All it does is creating a new thread that you could've created on your own and adds a ton of complexity to the "Server" class.

Also, I understand why it was previously a bit confusing for some people (blocking forever), however with this current PR, it seems to me that it may fix that confusion:

- If no condition passed - it's blocked forever (until stop is called)
- If there is a condition - it's blocked until the condition is met.

The `startAsNewThread` parameter here is redundant and blows up the complexity and the interface.

bamkrs: Ok, now I get it and I can understand your point. We can phase that out. I have an nice solution for this in my mind. I try to push it until tomorrow.

bamkrs:
I marked `run(bool)` as deprecated and the function prints a warning of its removal. The loop with the condition is implemented as separate function. 

I don't know why the Pipeline tests fail over and over again. Those weren't touched at all. However, sometimes they seem to work (as in the first push).

lganzzzo:
Hey @bamkrs ,

Looks good!

> I don't know why the Pipeline tests fail over and over again. Those weren't touched at all. However, sometimes they seem to work (as in the first push).

It looks that the test is failing to stop the server, and it's just hanging for 10 mins until the timeout.
The pipeline test is the first one that is using a test runner.
Seems like the `runConditionForLambda` is passed by value, thus the condition function always returns `false`.

bamkrs:
Oh snap! Can you remove those tests? I incidently pressed push instead of commit and now there is a bogous merge going on here...

bamkrs:
Whoops, somewhere down the line the correct call to the conditional function was lost. Now everything is in order and the test should run fine :)

> > I don't know why the Pipeline tests fail over and over again. Those weren't touched at all. However, sometimes they seem to work (as in the first push).
> 
> It looks that the test is failing to stop the server, and it's just hanging for 10 mins until the timeout.
> The pipeline test is the first one that is using a test runner.
> Seems like the `runConditionForLambda` is passed by value, thus the condition function always returns `false`.

Aaah, I assumed the test runner is used much earlier so I did not expect the error to be related to my changes. Quickly fixed everything. Would you (@lganzzzo) want to keep the small example in the test runner to test the proposed API or should I add an additional test? If so, there are no tests dedicated to the server iirc, do you have any good idea how to test it properly or should I just start a server, check that it runs, close it and check if it has stopped?

lganzzzo:
> Would you (@lganzzzo) want to keep the small example in the test runner to test the proposed API or should I add an additional test? If so, there are no tests dedicated to the server iirc, do you have any good idea how to test it properly or should I just start a server, check that it runs, close it and check if it has stopped?

I think we can keep the new functionality in the test runner and thus we have already this covered in the tests :)

bamkrs:
Any other suggestions, changes, comments?

src/oatpp/network/Server.hpp
```diff
@@ -111,12 +114,21 @@ class Server : public base::Countable {
...
+  void run(std::function<bool()> conditional = nullptr);
+
   /**
    * Call &id:oatpp::network::ConnectionProvider::getConnection; in the loop and passes obtained Connection
    * to &id:oatpp::network::ConnectionHandler;.
    * @param startAsNewThread - Start the server blocking (thread of callee) or non-blocking (own thread)
+   * @deprecated Deprecated since 1.3.0, will be removed in the next release.
    */
-  void run(bool startAsNewThread = false);
+  void run(bool startAsNewThread);
```
lganzzzo, line 131: I feel like we can drop it right away.
However, if you think it's better to keep it till the next release I'm ok with that too.

lganzzzo (approved):
Looks good!

src/oatpp/network/Server.hpp
bamkrs: Hm, this is a critical change in one of the core components. My guts tell me we should keep it and remove it in 1.3.0
However, I don't know if this feature is actually used that heavily. I want to ease the process of upgrading oat++ as much as possible. Therefore I think its our responsibility to warn users about breaking changes in minor releases (major releases are a whole other story). And since keeping this API alive until the next release does not hurt in any way (we don't have to implement compromises/adapters/workarounds) I assume this would be a nice move.

lganzzzo:
Great work!
Merged

PR #373, opened 2021-01-05
Fix wrong calling of conditional loop if condition is null (Closes #372)

### LINKED ISSUE #372: bad_function_call after calling server->run()
After updating to the most recent commit I'm getting a `bad_function_call` error on my project when calling `server->run()`.

Using 
```c++
std::function<bool()> condition = []{return true;};
  server->run(condition);
```
seems to fix it for now.

bamkrs:
If `server->run()` is called (no argument given, thus `conditional` is `nullptr`) the `conditionalMainLoop` was called instead of the normal `mainLoop`.
First reported in Issue #372.

Now the correct normal `mainLoop` is called when no conditional function was given.


lganzzzo:
Merged!

PR #378, opened 2021-01-12
Add optional default-value for QUERY parameters to ease the implementation of optional QUERY parameters

bamkrs:
This PR adds a fourth parameter to QUERY() where the string-representation of an default-value can be set. Thus allowing for optional QUERY parameters.


lganzzzo (approved):
Nice!

PR #380, opened 2021-01-14
Added operator= support for an std::string

sholmp:


src/oatpp/core/data/mapping/type/Primitive.hpp
```diff
@@ -110,6 +110,10 @@ class String : public type::ObjectWrapper<base::StrBuffer, __class::String> {
     return *this;
   }
 
+  inline String& operator = (const std::string& str) {
+    return operator=(str.c_str());
```
lganzzzo, line 114: This won't work for binary data.

You have to use the following instead:

```cpp
inline String& operator = (const std::string& str) {
  m_ptr = base::StrBuffer::createShared(str.data(), str.size(), true)
  return *this;
}
```

lganzzzo (changes requested):
To make this usable we have to add the following:

- Constructor from `std::string`
- Test - if `std::string` is passed to constructor/assign-operator the correct methods is used
- Test - if `const char*` param is passed to constructor/assign-operator the correct method is used. And that `const char*` doesn't get implicitly converted to `std::string` first.

src/oatpp/core/data/mapping/type/Primitive.hpp
sholmp: Hey @lganzzzo 
I assume these tests should be placed in "oatpp/test/oatpp/core/data/mapping/type/StringTest.cpp" right?
Also is there a way to easily run a single test? I can outcomment the others, but I would think you have a smarter way :slightly_smiling_face: ?

lganzzzo: Hey @sholmp,

Yes,- just comment out the other tests:)
I don't have a smarter solution for that.

bamkrs:
For further reference: Oat++ implements this feature natively at least since 247c994171432493b1b45ea7792edb8729f9275d

PR #386, opened 2021-01-30
Fix a 64bit function that was also being called when the environment …

Samega7Cattac:
…is 32btis.

The function _localtime64 was being called inside a ifdef WIN32 which
makes it try to compile even when it's in a 32bti environment.
Now if it calls _localtime32 in that case.
Tested on Windows using Msys+mingw 32bit shell.

Signed-off-by: Samega 7Cattac <7Cattac@gmail.com>


lganzzzo:
Great PR!
Merged!

PR #393, opened 2021-02-22
Let the ConnectionHandlers count how many threads they spawn.

bamkrs:
To reliable stop a ConnectionHandler it needs to count its spawned threads and wait for all threads to be done before stopping.
Because we can have multiple ConnectionHandlers, this counting can not be done via the `Components` since they *could* be shared between multiple ConnectionHandlers.
Thus we have to pass the counter separately.
While the counting can easily be done in the constructor and destructor of the Async-Api's `Coroutine` since it's threadpool cleanly allocs and deallocs the `Coroutine` the Simple-Api needs to be counted in the `Task::run()` method. It seems that `std::thread` does not deallocates the passed `HttpProcessor::Task` as expected (e.g. when the thread is done).

### DIFF
--- a/src/oatpp/web/server/AsyncHttpConnectionHandler.hpp
+++ b/src/oatpp/web/server/AsyncHttpConnectionHandler.hpp
@@ -38,7 +38,8 @@ class AsyncHttpConnectionHandler : public base::Countable, public network::Conne
 private:
   std::shared_ptr<oatpp::async::Executor> m_executor;
   std::shared_ptr<HttpProcessor::Components> m_components;
-  std::shared_ptr<std::atomic_ulong> m_spawns;
+  std::atomic_long m_spawns;
+  std::atomic_bool m_continue;
 public:
 
   /**
--- a/src/oatpp/web/server/AsyncHttpConnectionHandler.hpp
+++ b/src/oatpp/web/server/AsyncHttpConnectionHandler.hpp
@@ -37,8 +37,8 @@ namespace oatpp { namespace web { namespace server {
 class AsyncHttpConnectionHandler : public base::Countable, public network::ConnectionHandler {
 private:
   std::shared_ptr<oatpp::async::Executor> m_executor;
-private:
   std::shared_ptr<HttpProcessor::Components> m_components;
+  std::shared_ptr<std::atomic_ulong> m_spawns;
 public:
 
   /**
--- a/src/oatpp/web/server/HttpConnectionHandler.cpp
+++ b/src/oatpp/web/server/HttpConnectionHandler.cpp
@@ -39,6 +39,7 @@ namespace oatpp { namespace web { namespace server {
 
 HttpConnectionHandler::HttpConnectionHandler(const std::shared_ptr<HttpProcessor::Components>& components)
   : m_components(components)
+  , m_spawns(std::make_shared<std::atomic_ulong>(0))
 {}
 
 std::shared_ptr<HttpConnectionHandler> HttpConnectionHandler::createShared(const std::shared_ptr<HttpRouter>& router){
@@ -70,7 +71,7 @@ void HttpConnectionHandler::handleConnection(const std::shared_ptr<oatpp::data::
   connection->setInputStreamIOMode(oatpp::data::stream::IOMode::BLOCKING);
 
   /* Create working thread */
-  std::thread thread(&HttpProcessor::Task::run, HttpProcessor::Task(m_components, connection));
+  std::thread thread(&HttpProcessor::Task::run, HttpProcessor::Task(m_components, connection, m_spawns));
   
   /* Get hardware concurrency -1 in order to have 1cpu free of workers. */
   v_int32 concurrency = oatpp::concurrency::getHardwareConcurrency();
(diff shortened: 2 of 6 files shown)


src/oatpp/web/server/HttpProcessor.cpp
```diff
@@ -219,13 +219,17 @@ HttpProcessor::ConnectionState HttpProcessor::processNextRequest(ProcessingResou
...
+                          const std::shared_ptr<oatpp::data::stream::IOStream>& connection,
+                          const std::shared_ptr<std::atomic_ulong>& taskCounter)
   : m_components(components)
   , m_connection(connection)
+  , m_counter(taskCounter)
 {}
 
 void HttpProcessor::Task::run(){
 
+  (*m_counter)++;
```
lganzzzo, line 231: Should be incremented in the constructor of the `Task` or else,
someone can successfully call `stop()` before the thread started and `run()` was called.

src/oatpp/web/server/HttpProcessor.cpp
```diff
@@ -244,21 +248,31 @@ void HttpProcessor::Task::run(){
     // DO NOTHING
   }
 
+  (*m_counter)--;
```
lganzzzo, line 251: To my mind, a better place to dec counter is the destructor.

src/oatpp/web/server/HttpConnectionHandler.cpp
```diff
@@ -86,7 +87,21 @@ void HttpConnectionHandler::handleConnection(const std::shared_ptr<oatpp::data::
 }
 
 void HttpConnectionHandler::stop() {
```
lganzzzo, line 89: Once `stop()` was called, the `handleConnection()` should not populate any more threads.

src/oatpp/web/server/AsyncHttpConnectionHandler.hpp
```diff
@@ -37,8 +37,8 @@ namespace oatpp { namespace web { namespace server {
 class AsyncHttpConnectionHandler : public base::Countable, public network::ConnectionHandler {
 private:
   std::shared_ptr<oatpp::async::Executor> m_executor;
-private:
   std::shared_ptr<HttpProcessor::Components> m_components;
+  std::shared_ptr<std::atomic_ulong> m_spawns;
```
lganzzzo, line 41: Maybe it's better to store a plain atomic without shared_ptr wrapper - to avoid double atomic incs and decs
one for ptr counter and one for spawns

src/oatpp/web/server/HttpConnectionHandler.cpp
```diff
@@ -86,7 +87,21 @@ void HttpConnectionHandler::handleConnection(const std::shared_ptr<oatpp::data::
...
+  auto startTime = std::chrono::system_clock::now();
+  auto timeout = std::chrono::minutes(1);
+
+  while(m_spawns->load() != 0) {
+    auto elapsed = std::chrono::system_clock::now() - startTime;
+    if(elapsed < timeout) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    } else {
+      if (m_spawns->load() > 0) {
+        OATPP_LOGW("[oatpp::web::server::HttpConnectionHandler::stop()]", "Timeout while waiting 1 minute for all connections to close. There are %ul connections still alive.", m_spawns->load());
```
lganzzzo, line 100: let's just block forever - until all threads are done.
Exiting this loop having running server threads will put an application into an unpredictable state.

src/oatpp/web/server/HttpConnectionHandler.hpp
```diff
@@ -37,6 +37,7 @@ namespace oatpp { namespace web { namespace server {
 class HttpConnectionHandler : public base::Countable, public network::ConnectionHandler {
 private:
   std::shared_ptr<HttpProcessor::Components> m_components;
+  std::shared_ptr<std::atomic_ulong> m_spawns;
```
lganzzzo, line 40: Let's have it `signed` to avoid rounding in case of an error.
It also makes sense to add a redundant check after decrementing this value that it didn't go less than 0.

src/oatpp/web/server/HttpProcessor.cpp
bamkrs: Have you read the commit message? That is exactly I was doing in the first place. However, `std::thread` does not reliably clean the `HttpProcessor` on thread exit. It seems that the resources are only cleaned when the ConnectionHandler is cleaned.

src/oatpp/web/server/AsyncHttpConnectionHandler.hpp
bamkrs: I was thinking the same but couldn't decide between clean pointer passing or performance. But performance is key at this point, you are right.

src/oatpp/web/server/HttpConnectionHandler.cpp
bamkrs: Ok!

src/oatpp/web/server/HttpConnectionHandler.hpp
bamkrs: I couldn't possibly think of a scenario where this gets more decremented than incremented. Do you have a hint?

src/oatpp/web/server/HttpConnectionHandler.cpp
bamkrs: Oh yeah, will do!

src/oatpp/web/server/HttpConnectionHandler.hpp
bamkrs: > I couldn't possibly think of a scenario where this gets more decremented than incremented. Do you have a hint?

Except of the missing copy-constructor ;D

PR #395, opened 2021-02-28
Fix #324

### LINKED ISSUE #324: Windows mingw compile error on oatppAllTests project
ConnectionProvider.cpp:300:5: error: 'inet_ntop' was not declared in this scope inet_ntop(AF_INET, &sockAddress->sin_addr, strIp, INET_ADDRSTRLEN);
Windows, MinGW64, g++, cmake, clion

wvffle:
https://github.com/oatpp/oatpp/issues/324#issuecomment-711065560 describes how MinGW is failing to build due to its Windows XP support and this PR implements the failing function based on the code from https://www.mail-archive.com/users@ipv6.org/msg02107.html.

### DIFF
--- a/src/oatpp/network/tcp/server/ConnectionProvider.cpp
+++ b/src/oatpp/network/tcp/server/ConnectionProvider.cpp
@@ -43,6 +43,29 @@
   #endif
 #endif
 
+#if _WIN32_WINNT < 0x0600
+  const char * inet_ntop (int af, const void *src, char *dst, socklen_t cnt) {
+    if (af == AF_INET) {
+      struct sockaddr_in in;
+
+      memset (&in, 0, sizeof(in));
+      in.sin_family = AF_INET;
+      memcpy (&in.sin_addr, src, sizeof(struct in_addr));
+      getnameinfo ((struct sockaddr *)&in, sizeof (struct sockaddr_in), dst, cnt, NULL, 0, NI_NUMERICHOST);
+      return dst;
+    } else if (af == AF_INET6) {
+      struct sockaddr_in6 in;
+      memset (&in, 0, sizeof(in));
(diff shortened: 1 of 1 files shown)


src/oatpp/network/tcp/server/ConnectionProvider.cpp
```diff
@@ -43,6 +43,29 @@
   #endif
 #endif
 
+#if _WIN32_WINNT < 0x0600
```
bamkrs, line 46: Could you wrap the whole `#if` block in another check for an MinGW build like `#ifdef __MINGW32__
`? For linux compiler `#if _WIN32_WINNT < 0x0600` evaluates to true because the `_WIN32_WINNT` macro is not set.

src/oatpp/network/tcp/server/ConnectionProvider.cpp
```diff
@@ -43,6 +43,29 @@
...
+  const char * inet_ntop (int af, const void *src, char *dst, socklen_t cnt) {
```
bamkrs, line 47: Also, could you note where this code comes from? We should at least attribute the original author with a link to its original work.

bamkrs (changes requested):
Nice! Just two minor changes and I think we are ready to go!

lganzzzo (approved):
Looks good!

bamkrs (approved):
Nice!

bamkrs:
Good work @wvffle ! Now we can add MinGW in our compatibility list. This is a huge profit for Oat++ and a vast amount of potential users. Thank you.

wvffle:
> Good work @wvffle ! Now we can add MinGW in our compatibility list. This is a huge profit for Oat++ and a vast amount of potential users. Thank you for your contribution!

Glad to hear that :)

PR #396, opened 2021-03-01
Add hints and advice to the threading-examples in `Server.hpp`

bamkrs:
This PR adds some advice and hints to the threading-examples how to implement own threading when the deprecated threaded `run(true)` is removed.


lganzzzo (commented):
Look good!

Issue #403, opened 2021-03-23
Silently close connection if nullptr is returned on the endpoint


lganzzzo:

lganzzzo:
we are not going to implement this at the moment

PR #404, opened 2021-03-23
First step implementing silent connection drops. (#403)

bamkrs:
Server can now silently drop/close connections when an endpoint returns nullptr instead of an response. Tests are prepared but Oat++-client is not able to cope with it currently.
So still in need of improvements on the client side.

### DIFF
--- a/test/oatpp/web/FullAsyncTest.cpp
+++ b/test/oatpp/web/FullAsyncTest.cpp
@@ -187,6 +187,11 @@ void FullAsyncTest::onRun() {
         OATPP_ASSERT(dto);
         OATPP_ASSERT(dto->testValue == "my_test_header-Async");
       }
+
+      // { // test silent connection close
+      //   auto response = client->silentClose();
+      //   OATPP_ASSERT(response == nullptr);
+      // }
       
       { // test POST with body
         auto response = client->postBody("my_test_body-Async", connection);
--- a/test/oatpp/web/FullTest.cpp
+++ b/test/oatpp/web/FullTest.cpp
@@ -292,6 +292,11 @@ void FullTest::onRun() {
         OATPP_ASSERT(dto->testValue == "my_test_header");
       }
 
+      // { // test silent close
+      //   auto response = client->silentClose();
+      //   OATPP_ASSERT(response == nullptr);
+      // }
+
       { // test POST with body
         auto response = client->postBody("my_test_body", connection);
         OATPP_ASSERT(response->getStatusCode() == 200);
--- a/src/oatpp/web/server/HttpProcessor.cpp
+++ b/src/oatpp/web/server/HttpProcessor.cpp
@@ -155,6 +155,11 @@ HttpProcessor::ConnectionState HttpProcessor::processNextRequest(ProcessingResou
 
     response = processNextRequest(resources, request, connectionState);
 
+    // Silently close connection when endpoint returned a nullptr.
+    if (!response) {
+      return ConnectionState::CLOSING;
+    }
+
     try {
 
       for (auto& interceptor : resources.components->responseInterceptors) {
@@ -357,6 +362,10 @@ HttpProcessor::Coroutine::Action HttpProcessor::Coroutine::onRequestFormed() {
 }
 
 HttpProcessor::Coroutine::Action HttpProcessor::Coroutine::onResponse(const std::shared_ptr<protocol::http::outgoing::Response>& response) {
+  // Silently close connection when endpoint returned a nullptr.
+  if (!response) {
+    return finish();
+  }
   m_currentResponse = response;
   return yieldTo(&HttpProcessor::Coroutine::onResponseFormed);
 }
--- a/test/oatpp/web/app/Client.hpp
+++ b/test/oatpp/web/app/Client.hpp
@@ -55,6 +55,7 @@ public:
   API_CALL("GET", "queries/optional", getWithOptQueries, QUERY(String, name))
   API_CALL("GET", "queries/map", getWithQueriesMap, QUERY(String, key1), QUERY(Int32, key2), QUERY(Float32, key3))
   API_CALL("GET", "headers", getWithHeaders, HEADER(String, param, "X-TEST-HEADER"))
+  API_CALL("GET", "close", silentClose)
   API_CALL("POST", "body", postBody, BODY_STRING(String, body))
   API_CALL("POST", "body-dto", postBodyDto, BODY_DTO(Object<TestDto>, body))
 
@@ -92,6 +93,7 @@ public:
   API_CALL_ASYNC("GET", "queries", getWithQueriesAsync, QUERY(String, name), QUERY(Int32, age))
   API_CALL_ASYNC("GET", "queries/map", getWithQueriesMapAsync, QUERY(String, key1), QUERY(Int32, key2), QUERY(Float32, key3))
   API_CALL_ASYNC("GET", "headers", getWithHeadersAsync, HEADER(String, param, "X-TEST-HEADER"))
+  API_CALL_ASYNC("GET", "close", silentCloseAsync)
   API_CALL_ASYNC("POST", "body", postBodyAsync, BODY_STRING(String, body))
   API_CALL_ASYNC("POST", "echo", echoBodyAsync, BODY_STRING(String, body))
 
--- a/test/oatpp/web/app/Controller.hpp
+++ b/test/oatpp/web/app/Controller.hpp
@@ -141,7 +141,11 @@ public:
     dto->testValue = param;
     return createDtoResponse(Status::CODE_200, dto);
   }
-  
+
+  ENDPOINT("GET", "close", silentClose) {
+    return nullptr;
+  }
+
   ENDPOINT("POST", "body", postBody,
            BODY_STRING(String, body)) {
     //OATPP_LOGV(TAG, "POST body %s", body->c_str());
--- a/test/oatpp/web/app/ControllerAsync.hpp
+++ b/test/oatpp/web/app/ControllerAsync.hpp
@@ -90,7 +90,7 @@ public:
 
   ENDPOINT_ASYNC("GET", "headers", GetWithHeaders) {
 
-    ENDPOINT_ASYNC_INIT(GetWithHeaders)
+   ENDPOINT_ASYNC_INIT(GetWithHeaders)
 
     Action act() {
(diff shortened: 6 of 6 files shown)


test/oatpp/web/FullAsyncTest.cpp
```diff
@@ -187,6 +187,11 @@ void FullAsyncTest::onRun() {
         OATPP_ASSERT(dto);
         OATPP_ASSERT(dto->testValue == "my_test_header-Async");
       }
+
+      // { // test silent connection close
+      //   auto response = client->silentClose();
```
lganzzzo, line 192: Client should just throw an error in this situation (I mean it should be already working:)):

```cpp
try{
      auto response = client->silentClose();
      OATPP_ASSERT(response == nullptr);
} catch (const std::runtime_erro& e) {
   ...
}
```

test/oatpp/web/FullTest.cpp
```diff
@@ -292,6 +292,11 @@ void FullTest::onRun() {
         OATPP_ASSERT(dto->testValue == "my_test_header");
       }
 
+      // { // test silent close
```
lganzzzo, line 295: Client should just throw an error in this situation (I mean it should be already working:)):

```cpp
try{
      auto response = client->silentClose();
      OATPP_ASSERT(response == nullptr);
} catch (const std::runtime_erro& e) {
   ...
}
```

lganzzzo:
Also, there is a question about `Request Interceptors`:
Currently, to continue we return `nullptr`. So it conflicts with the new functionality.
Maybe we should introduce an additional flag in request interceptor method to designate that connection should be closed

bamkrs:
Yes, that collides with the implementation of the Interceptors. I forgot about that. Honestly, I dont like that passing of the special flag. Its something we dont do anywhere else and it feels so un-c++

I might have a different approach. Lets overload `Response`:

```
class NoResponse : public Response {
public:
  NoResponse() : Response(Status::NO_RESPONSE, nullptr) {}
}
```

So one could return this `NoResponse` response and we can check the body for `nullptr` in combination with a special `Status`. 
I don't see any issue with extending the `Status` class with special response-types since the naming scheme is already pretty clear: `Status::CODE_xyz` does already tells us "These are http-status-codes". Therefore we could extend `Status` with special states or "commands" like `Status::NO_RESPONSE`, `Status::RAW_RESPONSE` (for only sending the body, no headers, nothing, this might ease the implementation of other upgrade-handlers..?)

What do you think?

lganzzzo:
@bamkrs ,

> What do you think? There are so many options how to implement this.

Yes, so many options and It's hard to think of a good one:)

> Honestly, I dont like that passing of the special flag. Its something we dont do anywhere else and it feels so un-c++

Agree.

However, I also don't like the no-response response and the status `Status::NO_RESPONSE`,- It feels weird having to create a
`Response` object to designate the no-response. 
Also, 204 is not a `null`, it still has headers that you might have wanted to set.

I am currently thinking of the following:

So there is an expected flow:

- Endpoint - user must return a response.
- Interceptor - user must return a response or a `nullptr` designation an immediate response and a `continue` respectively.

If something goes wrong in the "expected flow" like the response is `null` on the endpoint, or it was an unexpected error during the DB access, or whatever else unexpected/unhandled situation - it's clearly a `500 - internal server error`.

Now the no-response, silent connection close feature:

```cpp
// For some reason I think that connection is invalid that's why I want to silently close it.
// Ex.: 
// - The peer is not expected to connect to my server - invalid connection
// - An invalid message received - invalid connection
// - An unexpected packet came in - invalid connection
throw InvalidConnectionError();
```

What do you think?

bamkrs:
> However, I also don't like the no-response response and the status `Status::NO_RESPONSE`,- It feels weird having to create a
> `Response` object to designate the no-response.

Hm, I know, it sounds weird to create a special "no-response"-response but it would be a clean solution workflow-wise and we could use the same technique with a "continue" response for interceptors. It would streamline the internal APIs

> Also, 204 is not a `null`, it still has headers that you might have wanted to set.

I think you misunderstood me here. I thought about treating a `return nullptr` differently. The 204 has a empty body. So it *could* be one of the many options what Oat++ should do when an endpoint `return nulltpr`.

> I am currently thinking of the following:
> 
> So there is an expected flow:
> 
>     * Endpoint - user must return a response.
> 
>     * Interceptor - user must return a response or a `nullptr` designation an immediate response and a `continue` respectively.
> 
> 
> If something goes wrong in the "expected flow" like the response is `null` on the endpoint, or it was an unexpected error during the DB access, or whatever else unexpected/unhandled situation - it's clearly a `500 - internal server error`.
> 
> Now the no-response, silent connection close feature:
> 
> ```c++
> // For some reason I think that connection is invalid that's why I want to silently close it.
> // Ex.: 
> // - The peer is not expected to connect to my server - invalid connection
> // - An invalid message received - invalid connection
> // - An unexpected packet came in - invalid connection
> throw InvalidConnectionError();
> ```
> 
> What do you think?

I don't like using Exceptions as means of control flow. A decision to not answer isn't exclusively an error condition and throwing Exceptions to tell Oat++ what to do feels weirder then returning a "no response"-response.

---

Lets rewind and start here again:

> I am currently thinking of the following:
> 
> So there is an expected flow:
> 
>     * Endpoint - user must return a response.
> 
>     * Interceptor - user must return a response or a `nullptr` designation an immediate response and a `continue` respectively.

When we reword this workflow and reset our heads, returning `nullptr` as "no-response" in an Endpoint might be a very streamlined decision:

An interceptor is expected to return a response, if it doesn't the next interceptor is called and so on. Lets treat the endpoint as just another link in the request-chain!
So compared to the original proposal, where the calling-chain ends when the endpoint returns `nullptr`, I would now propose that after the `nullptr` from the endpoint the response interceptors are still called but have to be prepared to handle a `nullptr` as parameter.
Only at the end of our RQInterceptor-Endpoint-RSPInterceptor we evaluate if we have an valid response or still have a `nullptr`. With a response we respond, with `nullptr` we just close the connection.

In my head, this is a perfectly logical workflow. It would also have the charm that the user could still do some special stuff in the response-interceptors when the endpoint returns a `nullptr`.

Does this workflow feels more natural to you?

lganzzzo:
@bamkrs ,

> I don't like using Exceptions as means of control flow. A decision to not answer isn't exclusively an error condition and throwing Exceptions to tell Oat++ what to do feels weirder then returning a "no response"-response.

Not really - since I'm thinking in terms of HTTP (and these changes will affect exactly HTTPProcessor).
The server is always required to answer to client's request. The only reason why a client can get no answer - is a connection error - in which case - it's ok to drop a connection.

> An interceptor is expected to return a response, if it doesn't the next interceptor is called and so on. Lets treat the endpoint as just another link in the request-chain!

Let's consider a blacklisted-IPs scenario.
In this case, it makes sense to implement a request interceptor which, if spotted a malicious activity, will add clients' IP to a blacklist and will drop connection immediately.
So the request processing chain should be broken.

> So compared to the original proposal, where the calling-chain ends when the endpoint returns nullptr, I would now propose that after the nullptr from the endpoint the response interceptors are still called but have to be prepared to handle a nullptr as parameter.

It's a good point - whether or not `nullptr` response should be handed to response interceptors.
What is clear to me, is that in case of a `nullptr` response, the error should be reported (same way as in the response interceptors' chain - https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/server/HttpProcessor.cpp#L162).
The question is - should we generate an error-response, and then hand it to response interceptors, or should we report straight to a client skipping response interceptors.

lganzzzo:
The `ControlFlow` feature is definitely faster than throwing an exception. But unfortunately, we'll have to put it in every place where we deal with a response:

```cpp
/* In ENDPOINT */
ENDPOINT(...) {
  auto response = createResponse(...);
  response->putHeader(...);
  return ControlFlow<std::shared_ptr<OutgoingResponse>>(response); 
}

/* In response interceptor */
class MyResponseInterceptor : ... {
public:
  ControlFlow<std::shared_ptr<OutgoingResponse>> intercept(const std::shared_ptr<IncomingRequest>& request,
                                                           const ControlFlow<std::shared_ptr<OutgoingResponse>>& response)
  {
    return response;
  }

}
```

---

> Okay, I might sense a conflict of interests or very different opinions. I understand your point and try to find a solution which is acceptable for both of us.

Well, it looks that we are back to where we started - oatpps' REST framework wasn't really designed for no-responses.

I'm viewing it as follows:

Since "no-responses" is not something that oatpp was originally designed for, we shouldn't sacrifice the current design to "force-push" this feature.
So I thought: 
 - Ok, currently we are handling a `nullptr` response in the response interceptor by generating a [`500` error response](https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/server/HttpProcessor.cpp#L162) so it will be logical to do the same if an endpoint returns `nullptr`.
 - Ok, we are already using exceptions to report errors, plus we are [terminating connection if there is an IO error](https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/server/HttpProcessor.cpp#L138) - so why not introduce a new specific  `FatalConnectionError` to do the same. The implementation will be as easy as follows by adding a catch clause [here](https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/server/HttpProcessor.cpp#L123v):

```cpp
  ...
} catch (FatalConnectionError e) {
  connectionState = ConnectionState::DEAD;
}
```

Thus no of the current API/behavior changed:
- RequestInterceptor works as before - `return nullptr` == continue.
- Endpoint works as before, except that now it reports `500` instead of a seg-falt on a `nullptr` response.
- ResponseInterceptor works as before - no changes.
- A new specific exception class is added designating a fatal connection error.

What do you think?

bamkrs:
So I did implement the special exception, however the endpoint can still return `nullptr` to drop the connection. I can still change that. 
However I did run into another problem: While the method works for virtual connections, on "real" connections the client seems to return a valid response even when the controller just closes the connection. I'm not fluent enough in the client code that I could debug that on the fly. Do you have any hints where I have to look?

lganzzzo:
> While the method works for virtual connections, on "real" connections the client seems to return a valid response even when the controller just closes the connection

Interesting... I thought it would lead to the exception - https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/client/HttpRequestExecutor.cpp#L141

We should check what [headerReader.readHeaders()](https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/client/HttpRequestExecutor.cpp#L131) returns. If it returns some garbage - then it must be a bug.

PR #414, opened 2021-04-22
Update port after binding

boldrij:
Typicaly in case we use a port equal 0 to let the system find a free port, we need to update the port property of the provider


lganzzzo:
Looks good!
Thanks for the PR!

boldrij:
Super, thanks
Julien

PR #419, opened 2021-05-02
Added missing throw keyword

MHaselmaier:
Hi,

while browsing your source code I found these two functions which create a new std::runtime_error on the stack, but never throw it.
I think the throw keyword was forgotten.

PS: This is my first ever contribution to open source projects. If I overlooked something in the workflow, please let me know.


lganzzzo:
Hey @MHaselmaier ,

You are right, this is a missing `throw`.

Thanks for fixing it!

PR #420, opened 2021-05-02
Minor refactorings

MHaselmaier:
Hi,

I made some small refactorings.
In my opinion the added override specifiers are especially important.

### DIFF
--- a/src/oatpp/core/data/stream/BufferStream.cpp
+++ b/src/oatpp/core/data/stream/BufferStream.cpp
@@ -147,7 +147,7 @@ oatpp::async::CoroutineStarter BufferOutputStream::flushToStreamAsync(const std:
       , m_stream(stream)
     {}
 
-    virtual Action act() override {
+    Action act() override {
       if(m_inlineData.currBufferPtr == nullptr) {
         m_inlineData.currBufferPtr = m_this->m_data;
         m_inlineData.bytesLeft = m_this->m_position;
--- a/src/oatpp/core/data/stream/BufferStream.cpp
+++ b/src/oatpp/core/data/stream/BufferStream.cpp
@@ -147,7 +147,7 @@ oatpp::async::CoroutineStarter BufferOutputStream::flushToStreamAsync(const std:
       , m_stream(stream)
     {}
 
-    Action act() {
+    virtual Action act() override {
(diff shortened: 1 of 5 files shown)


src/oatpp/core/data/stream/BufferStream.cpp
```diff
@@ -147,7 +147,7 @@ oatpp::async::CoroutineStarter BufferOutputStream::flushToStreamAsync(const std:
       , m_stream(stream)
     {}
 
-    Action act() {
+    virtual Action act() override {
```
lganzzzo, line 150: Please use just `override`. No need in `virtual` here

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -119,7 +119,7 @@ async::CoroutineStarter WriteCallback::writeExactSizeDataAsync(const void* data,
       , m_inlineData(data, size)
     {}
 
-    Action act() {
+    virtual Action act() override {
```
lganzzzo, line 122: Please use just `override`. No need in `virtual` here

src/oatpp/web/protocol/http/outgoing/Request.cpp
```diff
@@ -165,7 +165,7 @@ oatpp::async::CoroutineStarter Request::sendAsync(std::shared_ptr<Request> _this
       , m_headersWriteBuffer(std::make_shared<oatpp::data::stream::BufferOutputStream>())
     {}
     
-    Action act() {
+    virtual Action act() override {
```
lganzzzo, line 168: Please use just `override`. No need in `virtual` here

lganzzzo (approved):
Looks good!
Please address the minor comments.

MHaselmaier:
I used both virtual and override because it was use in this way on other classes.
Would you be interessted in a global change to only use one of the virtual, override and final specifiers?
I could post another pull request, it should be an easy task with tool assistence.

lganzzzo:
> Would you be interessted in a global change to only use one of the virtual, override and final specifiers?

Sounds good - we really need some code cleanup.
Also, please do not put final specifiers, since we want most of the APIs to be overridable.

lganzzzo:
Merged!

PR #423, opened 2021-05-03
Fixed memory leak in move assignement operators

MHaselmaier:
Since the destructors call the delete operator, it should also be called in the move assignement operators.


lganzzzo:
Great fix!
Merged

PR #426, opened 2021-05-06
Fixed compiler warnings

MHaselmaier:
When comiling the project with MSVC I got a few warnings for truncations and unreferenced variables.
Clang/GCC also produce the truncation warnings if the -Wconversion flag is set.


lganzzzo:
Nice!
Merged

PR #435, opened 2021-06-10
Add option to link with MSVC static runtime (OFF by default)

sgallou:
Hello,

this PR add the possibility to link oatpp with static runtime library (/MT, /MTd) under MSVC compiler when using CMake.
To use it, just set the 'OATPP_MSVC_LINK_STATIC_RUNTIME' option to ON (OFF by default to not change current behaviour).

Let me know if you prefer I move the cmake macro 'configure_msvc_static_runtime' into the existing CMakeLists.txt file instead of dedicated a new dedicated 'MSVCStaticRuntime.cmake' file, or for any other change.

I hope this improvement will be helpful.


src/CMakeLists.txt
```diff
@@ -278,6 +278,11 @@ set_target_properties(oatpp PROPERTIES
         CXX_STANDARD_REQUIRED ON
 )
 
+if(OATPP_MSVC_LINK_STATIC_RUNTIME)
```
lganzzzo, line 281: Let's move this block to the `root/CMakeLists.txt`

src/MSVCStaticRuntime.cmake
```diff
@@ -0,0 +1,27 @@
+macro(configure_msvc_static_runtime)
```
lganzzzo, line 1: Let's move this file to `root/cmake/MSVCStaticRuntime.cmake`

lganzzzo (commented):
Looks good!
Please see minor comments

sgallou:
In fact, CMake cache was not well manage. I fixed it in this commit.
And this setting was not applied to all projects (ie oatpp-test), it's also fixed.
And I tried to take account of your comments.

lganzzzo:
Great PR!
Merged

PR #450, opened 2021-07-29
Catching up with streaming changes

elgatito:
Hello,

Creating this PR to not forget about the changes.

There were 2 issues:
1) Having int64 in getKnownSize to properly work with big values in Content-Length.
2) Properly check for getKnownSize to show Content-Length (if available), instead of Transfer-Encoding.

You can, of course, close the PR and add changes on your own. 
I would prefer using non-forked repo. 

Thanks


lganzzzo:
Merged!

PR #464, opened 2021-08-17
Streams: `FIFOStream` implementation and `BufferedInputStream` API

bamkrs:
Oat++ Streams are great. However, I always missed an easy to use Stream-Buffer where we can write to AND read from.
`FIFOStream` aims to archive this. Using `oatpp::data::buffer::FIFOBuffer` we now have the ability to have an buffered write-callback enabled `InputStream`.

Also `BufferedInputStream` was introduced. It should unify the API's used for the several buffered `InputStream` implementations.

### DIFF
--- a/changelog/1.3.0.md
+++ b/changelog/1.3.0.md
@@ -54,12 +54,32 @@ Now it's much easier to use `oatpp::String` since `oatpp::String` is now wrapper
 [#408](https://github.com/oatpp/oatpp/issues/408)
 
 ```cpp
-auto connectionPool = std::make_shared<ClientConnectionPool>(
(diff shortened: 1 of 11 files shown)


lganzzzo:
> Also BufferedInputStream was introduced. It should unify the API's used for the several buffered InputStream implementations.

This one looks nice. I will review this PR in the evening.

For some reason both your PRs include changes that are already merged to `1.3.0`.
Please try to resolve that if possible.

bamkrs:
Yeah, I really don't know whats happening. I freshly pulled oatpp from github. Checked out v1.3.0, created a new branch and added the changes. Something is very odd.

PR #466, opened 2021-08-17
Streams: New `FIFOStream` stream and `BufferedInputStream` interface

bamkrs:
Oat++ Streams are great! However, I always missed a buffered stream that acts as `InputStream` with a writable buffer. `InputBufferStream` only works with an static memory.
To solve this `FIFOStream` is an `InputStream` that also has an write-callback to dynamically insert more data to the buffer.
As a bonus, it is resizeable.

Now we have several buffered streams that somehow expose similar APIs that kind of do the same. `BufferedInputStream` is a new interface that unifies those different APIs and enables generalization.


bamkrs:
I will add the corresponding notes to the changelog after the core-review  :)

lganzzzo (approved):
Looks good!

PR #468, opened 2021-08-19
`FIFOBuffer` should accept an `WriteCallback`/`ReadCallback`

bamkrs:
`FIFOBuffer` should accept an `WriteCallback`/`ReadCallback` to its `readAndWriteToStream` and `readFromStreamAndWrite` functions.

There is no need to specialize on a `Stream`.

Its an non-breaking change. Also renaming the functions to `readAndWriteToCallback` and `readFromCallbackAndWrite` would break things.


lganzzzo (approved):

lganzzzo:
Merged!

PR #469, opened 2021-08-19
WIP: Stream API rework

bamkrs:
Following my last stream API PR, this one build upon the concepts of the first PR and majorly touches all streams that somehow use a buffer. 
Again, all `OutputStream` had a similiar API with similar features. Namely the `flushToStream` function. However, those streams did not inherit from an common base class so it is hard to use different implementations in a general context. 

The new `BufferedStream` interface packs a set of functions that were found on all buffered streams (Input and Output) and provides default implementations using `stream::transfer`.
Those default implementations can be overridden to maximize performance like in `FIFOStream` which directly writes to the next streams `WriteCallback` instead of using `stream::transfer` which requires an intermediate buffer.

For input-streams, the implementation is quite trivial and builds upon already existing APIs. For output-streams a lot of specialized implementation is needed. 

The PR introduces breaking changes to the current API and is still a WIP. 

ToDo:

- [ ]  `writeBufferToStream` with read-offsets. 
- [ ]  Introduce better names for functions

### DIFF
--- a/src/oatpp/core/data/stream/BufferStream.cpp
+++ b/src/oatpp/core/data/stream/BufferStream.cpp
@@ -128,37 +128,56 @@ oatpp::String BufferOutputStream::getSubstring(v_buff_size pos, v_buff_size coun
   }
 }
 
-oatpp::v_io_size BufferOutputStream::flushToStream(OutputStream* stream) {
+v_io_size BufferOutputStream::flushBufferToStream(stream::WriteCallback* stream) {
   return stream->writeExactSizeDataSimple(m_data, m_position);
 }
 
-oatpp::async::CoroutineStarter BufferOutputStream::flushToStreamAsync(const std::shared_ptr<BufferOutputStream>& _this, const std::shared_ptr<OutputStream>& stream) {
+oatpp::async::CoroutineStarter BufferOutputStream::flushBufferToStreamAsync(const std::shared_ptr<BufferOutputStream>& _this, const std::shared_ptr<WriteCallback>& stream) {
+  return writeBufferToStreamAsync(_this, stream, _this->m_position);
+}
 
+oatpp::async::CoroutineStarter BufferOutputStream::writeBufferToStreamAsync(const std::shared_ptr<BufferOutputStream> &_this,
+                                                                            const std::shared_ptr<WriteCallback> &stream,
+                                                                            v_buff_size count) {
   class WriteDataCoroutine : public oatpp::async::Coroutine<WriteDataCoroutine> {
-  private:
+   private:
     std::shared_ptr<BufferOutputStream> m_this;
-    std::shared_ptr<oatpp::data::stream::OutputStream> m_stream;
+    std::shared_ptr<oatpp::data::stream::WriteCallback> m_stream;
     data::buffer::InlineWriteData m_inlineData;
-  public:
+    v_buff_size m_count;
+   public:
 
     WriteDataCoroutine(const std::shared_ptr<BufferOutputStream>& _this,
-                       const std::shared_ptr<oatpp::data::stream::OutputStream>& stream)
-      : m_this(_this)
-      , m_stream(stream)
+                       const std::shared_ptr<oatpp::data::stream::WriteCallback>& stream,
+                       v_buff_size count)
+        : m_this(_this)
+        , m_stream(stream)
+        , m_count(count)
     {}
 
     Action act() override {
       if(m_inlineData.currBufferPtr == nullptr) {
         m_inlineData.currBufferPtr = m_this->m_data;
-        m_inlineData.bytesLeft = m_this->m_position;
+        m_inlineData.bytesLeft = std::min(m_this->m_position, m_count);
       }
       return m_stream.get()->writeExactSizeDataAsyncInline(m_inlineData, finish());
     }
 
   };
+  return WriteDataCoroutine::start(_this, stream, count);
+}
 
-  return WriteDataCoroutine::start(_this, stream);
+v_io_size BufferOutputStream::writeBufferToStream(stream::WriteCallback *writeCallback, v_buff_size count) {
+  return writeCallback->writeSimple(m_data, std::min(m_position, count));
+}
 
+async::CoroutineStarter BufferOutputStream::writeBufferToStreamAsync(const std::shared_ptr<data::stream::WriteCallback> &stream,
+                                                                     v_buff_size count) {
+  return BufferOutputStream::writeBufferToStreamAsync(shared_from_this(), stream, count);
+}
+
+v_buff_size BufferOutputStream::getSize() {
+  return m_position;
 }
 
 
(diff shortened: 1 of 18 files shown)


src/oatpp/core/data/stream/BufferStream.cpp
```diff
@@ -128,39 +128,87 @@ oatpp::String BufferOutputStream::getSubstring(v_buff_size pos, v_buff_size coun
...
+oatpp::async::CoroutineStarter BufferOutputStream::writeBufferToStreamAsync(const std::shared_ptr<BufferOutputStream> &_this,
+                                                                            const std::shared_ptr<WriteCallback> &stream,
+                                                                            v_buff_size count) {
+  return writeBufferToStreamAsync(_this, stream, count, 0);
+}
 
+oatpp::async::CoroutineStarter BufferOutputStream::writeBufferToStreamAsync(const std::shared_ptr<BufferOutputStream> &_this,
+                                                                            const std::shared_ptr<WriteCallback> &stream,
+                                                                            v_buff_size count, v_buff_size readOffset) {
   class WriteDataCoroutine : public oatpp::async::Coroutine<WriteDataCoroutine> {
```
lganzzzo, line 148: Looks like `WriteDataCoroutine` can be rewritten to something like this:

```cpp
  class WriteDataCoroutine : public oatpp::async::Coroutine<WriteDataCoroutine> {
   private:
    std::shared_ptr<BufferOutputStream> m_this;
    std::shared_ptr<oatpp::data::stream::WriteCallback> m_stream;
    data::buffer::InlineWriteData m_inlineData;
    v_buff_size m_count;
    v_buff_size m_readOffset;
   public:

    WriteDataCoroutine(const std::shared_ptr<BufferOutputStream>& _this,
                       const std::shared_ptr<oatpp::data::stream::WriteCallback>& stream,
                       v_buff_size count, v_buff_size readOffset)
        : m_this(_this)
        , m_stream(stream)
        , m_count(count)
        , m_readOffset(readOffset)
    {
      m_inlineData.set(m_this->m_data + m_readOffset, std::min(m_this->m_position - m_readOffset, m_count));
    }

    Action act() override {
      return m_stream->writeExactSizeDataAsyncInline(m_inlineData, finish());
    }

  };
```

There is no need in additional checks, plus inline data initialization can be moved to constructor.

Also, if `if (m_readOffset > m_this->m_position) ` then it's better to not start the coroutine:

```cpp
if(readOffset < m_position) {
  return WriteDataCoroutine::start(_this, stream, count, readOffset);
}
return nullptr; // do not start the coroutine
```

src/oatpp/core/data/stream/ChunkedBuffer.hpp
```diff
@@ -36,7 +36,7 @@ namespace oatpp { namespace data{ namespace stream {
 /**
  * Buffer wich can grow by chunks and implements &id:oatpp::data::stream::ConsistentOutputStream; interface.
  */
-class ChunkedBuffer : public oatpp::base::Countable, public ConsistentOutputStream, public std::enable_shared_from_this<ChunkedBuffer> {
+class ChunkedBuffer : public oatpp::base::Countable, public BufferedOutputStream, public std::enable_shared_from_this<ChunkedBuffer> {
```
lganzzzo, line 39: I propose to consider dropping `ChunkedBuffer` class and replace it completely with `BufferOutputStream`

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -284,6 +284,82 @@ StreamType DefaultInitializedContext::getStreamType() const {
...
+v_io_size BufferedInputStream::flushBufferToStream(stream::WriteCallback *writeCallback) {
+  return writeBufferToStream(writeCallback, availableToRead());
+}
+
+async::CoroutineStarter BufferedInputStream::flushBufferToStreamAsync(const std::shared_ptr<data::stream::WriteCallback> &writeCallback) {
+  return writeBufferToStreamAsync(writeCallback, availableToRead());
+}
+
+v_io_size BufferedInputStream::writeBufferToStream(stream::WriteCallback *writeCallback, v_buff_size count) {
+  auto ioBuffer = buffer::IOBuffer::createShared();
```
lganzzzo, line 299: In simple API methods, there is no need for a shared object.
It might be better to use a regular `std::unique_ptr<char[]> buffer(new char[<size of IOBuffer>])`

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -284,6 +284,82 @@ StreamType DefaultInitializedContext::getStreamType() const {
...
+
+async::CoroutineStarter BufferedInputStream::writeBufferToStreamAsync(const std::shared_ptr<data::stream::WriteCallback>& writeCallback, v_buff_size count) {
+  auto ioBuffer = buffer::IOBuffer::createShared();
+  return transferAsync(this, writeCallback, count, ioBuffer);
+}
+
+v_io_size BufferedInputStream::writeBufferToStream(stream::WriteCallback *writeCallback,
+                                                   v_buff_size count,
+                                                   v_buff_size readOffset) {
+  auto ioBuffer = buffer::IOBuffer::createShared();
```
lganzzzo, line 313: let's allocate resources after the condition

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -284,6 +284,82 @@ StreamType DefaultInitializedContext::getStreamType() const {
...
+  transfer(this, writeCallback, count, ioBuffer->getData(), ioBuffer->getSize());
+  return count;
+}
+
+async::CoroutineStarter BufferedInputStream::writeBufferToStreamAsync(const std::shared_ptr<data::stream::WriteCallback> &stream,
+                                                                      v_buff_size count,
+                                                                      v_buff_size readOffset) {
+  auto ioBuffer = buffer::IOBuffer::createShared();
+  if (readOffset > availableToRead()) {
+    class NothingToDo : public async::Coroutine<NothingToDo> {
```
lganzzzo, line 329: `NothingToDo` coroutine is not needed.
` return nullptr;` will do

src/oatpp/core/data/stream/ChunkedBuffer.hpp
bamkrs: Do you mean to drop `ConsistentOutputStream` and refactor its functions to `BufferOutputStream`? It has all those `writeAsString(xxx)` helper functions.

Thats what I was thinking to do in the first place but found that you had a special use-case for `ConsistentOutputStream` and did not want to touch it before you reviewing.

src/oatpp/core/data/stream/Stream.cpp
bamkrs: TIL. Thats good to know!

src/oatpp/core/data/stream/BufferStream.cpp
bamkrs: Yes, now that I have some experience with `oatpp::async` I would have written it differently in the first place.

src/oatpp/core/data/stream/ChunkedBuffer.hpp
lganzzzo: No, just to drop `ChunkedBuffer `.
`ConsistentOutputStream` is needed for serializer. Serializers work only with `ConsistentOutputStream` at the moment.

Issue #470, opened 2021-08-19
oatpp::String / oatpp::base::StrBuffer interoperable with std::string


grisuthedragon:
Is there a way to get an easy way in both directions between `std::string` and Oat++'s String/StrBuffer types. Using 
```
 str->std_str() 
```
or 
```
 String::createShared(sstr.c_str()) 
``` 
at every point creates a lot of bloat in the code when dealing with other libraries, that relying on the `std::string` type. 

If there is no way, one can see this as feature request. I would be find if 
```
std::string sstr("test");
auto ostr = oatpp::String::createShared(sstr); 
``` 
and 
```
auto ostr = oatpp:String("mystring");
std::string sstr = ostring;
```
would work. That you make huge parts of our code better readable.

bamkrs:
Hey Martin,

that's one of the most common Problems with Oat++ strings, not gonna lie 👀
Have you checked the v1.3.0 branch? It's our next release. One of its mayor changes include the refactoring `oatpp::String` to be 'just a wrapper' around `std::string`. 
This highly reduces the bloat and prevents unnecessary copying of data.

grisuthedragon:
Basically this looks good but having the `*` in one direction is a bit strange. So having the proper type cast function in this place would be great. Another problem is, that this breaks the swagger support at the moment.

bamkrs:
Hey Martin,

you are welcome to contribute a pull request, we always encourage our community to provide their own PR's so its openly documented where the suggestions comes from :)
If you would rather not, we take care.

Also, swagger does have an `1.3.0` branch, too. Is it broken? Haven't checked in a while and `1.3.0` introduces some major refactoring of core components.

grisuthedragon:
I added the patch, including the requirements for the `oatpp-swagger` module.  The a separate PR https://github.com/oatpp/oatpp-swagger/pull/50 is placed to let it work again.

lganzzzo:
Fixed in `1.3.0` - see [the new oatpp::String](https://github.com/oatpp/oatpp/blob/master/changelog/1.3.0.md#the-new-oatppstring)

PR #471, opened 2021-08-24
oatpp::String works better with std::string

grisuthedragon:
As suggested in #470, the typecast operator from `oatpp::String` to `std::string` is included. 

Furthermore, to get `oatpp-swagger` running again, a case insensitive string comparison is required. For this purpose the `oatpp::String` class gets the following new methods:
```cpp
inline bool equalsCI(const std::string &sb) ;
inline bool equalsCI(const String &b);
inline bool equalsCI(const char *b);
```

Finally, trialling whitespaces are removed.

Closes: #470

### DIFF
--- a/src/oatpp/core/data/mapping/type/Primitive.hpp
+++ b/src/oatpp/core/data/mapping/type/Primitive.hpp
@@ -32,6 +32,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <iterator>
 
 namespace oatpp { namespace data { namespace mapping { namespace type {
 
@@ -128,7 +129,8 @@ public:
 
   operator std::string() const
   {
-      return this->m_ptr.operator*();
+    if (this->m_ptr == nullptr) throw std::runtime_error("m_ptr points to null.");
+    return this->m_ptr.operator*();
   }
 
   template<typename T,
@@ -175,6 +177,7 @@ public:
 
 
   inline bool equalsCI(const std::string &sb) {
+    if (this->m_ptr == nullptr ) return false;
     const std::string& sa = this->m_ptr.operator*();
     return (sa.size() == sb.size()) && std::equal(sa.begin(), sa.end(), sb.begin(),
                       [] (char a, char b) -> bool {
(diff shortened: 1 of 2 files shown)


src/oatpp/core/data/mapping/type/Primitive.hpp
```diff
@@ -123,6 +126,11 @@ class String : public type::ObjectWrapper<std::string, __class::String> {
     return this->m_ptr.operator*();
   }
 
+  operator std::string() const
```
lganzzzo, line 129: Since it's an implicit conversion with implicit dereferencing
I suggest adding `throw` in case of `m_ptr == nullptr`

src/oatpp/core/data/mapping/type/Primitive.hpp
```diff
@@ -165,6 +173,25 @@ class String : public type::ObjectWrapper<std::string, __class::String> {
     return *this;
   }
 
+
+  inline bool equalsCI(const std::string &sb) {
+    const std::string& sa = this->m_ptr.operator*();
```
lganzzzo, line 178: Should return `false` in case `m_ptr == nullptr`

src/oatpp/core/data/mapping/type/Primitive.hpp
```diff
@@ -165,6 +173,25 @@ class String : public type::ObjectWrapper<std::string, __class::String> {
...
+                          return std::tolower(a) == std::tolower(b);
+                      });
+  }
+
+  inline bool equalsCI(const String &b) {
+    const std::string& sb = *b;
+    return this->equalsCI(sb);
+  }
+
+  inline bool equalsCI(const char *b) {
```
lganzzzo, line 190: Should return `true` if both `this->m_ptr == nullptr` and `b == nullptr`
Should return `false` if `this->m_ptr != nullptr` and `b == nullptr`
Should return `false` if `this->m_ptr == nullptr` and `b != nullptr`

src/oatpp/core/data/mapping/type/Primitive.hpp
```diff
@@ -165,6 +173,25 @@ class String : public type::ObjectWrapper<std::string, __class::String> {
...
+    const std::string sb(b);
```
lganzzzo, line 191: Should be possible to perform comparison without creating a new `std::string`

lganzzzo:
Hello @grisuthedragon ,

Thank you for the PR!
The suggested convenience API looks reasonable!

Please address review comments.

Regards,
Leonid

src/oatpp/core/data/mapping/type/Primitive.hpp
grisuthedragon: What should be thrown in this case? Would `std::runtime_error` be fine?

src/oatpp/core/data/mapping/type/Primitive.hpp
grisuthedragon: Should be fine now.

grisuthedragon (commented):
all things should be fine now,

src/oatpp/core/data/mapping/type/Primitive.hpp
bamkrs: Yes, `std::runtime_error` is the common ground we chose. Just take a look at the other exceptions. They are most of the time

`std::runtime_error("[<full qualified classname with namespace>] Error: <message>")`

grisuthedragon: Done.

lganzzzo:
Merged
Thanks for the PR!

PR #479, opened 2021-09-07
Try to fix cross compilation for Raspberry/Synology

sgallou:
Hi,

It seems that GCC doesn't support well initializer for ParamsTypeMap ([see this build](https://github.com/Yadoms/yadoms/runs/3535646691?check_suite_focus=true)), error was something like :
`
/work/projects/external-libs/oatpp/src/oatpp_build/src/oatpp/orm/Executor.cpp:42:64: error: converting to 'const ParamsTypeMap {...}' from initializer list would use explicit constructor '...'
`

This PR suggest a simple workaround using default ctor of ParamsTypeMap instead of empty initializer at parseQueryTemplate call.


lganzzzo:
Hey @sgallou ,

Thanks for the PR. I've restarted the build

lganzzzo:
Merged!

PR #482, opened 2021-09-10
Add API to replace headers in requests and responses (Closes #481).

### LINKED ISSUE #481: Overriding default Content-Type header results in duplicate values
I would like to override the default Content-Type header in the response.
I added putHeader, but the response has an extra "application/json". 

```
auto dto = TestDto::createShared(); 
dto->msg = "Hello World!";
auto response = createDtoResponse(Status::CODE_200, dto);
response->putHeader(Header::CONTENT_TYPE, "application/json;version=2");
return response;
```

Actual Response:
`content-type: application/json;version=2,application/json`

Expected Response:
`content-type: application/json;version=2`

bamkrs:


lganzzzo (commented):
Nice set of new APIs.

Also, it's missing the actual fix of (#481).

In order to fix it, we have to change `headers.put(Header::CONTENT_TYPE, m_contentType);` to `headers.putIfNotExists(Header::CONTENT_TYPE, m_contentType);`
here - https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/protocol/http/outgoing/BufferBody.cpp#L64

---

What if we rename `replaceOrPut` --> `putOrReplace` - this way all `put` related methods would start with `put` word

bamkrs:
> Nice set of new APIs.
> 
> Also, it's missing the actual fix of (#481).
> 
> In order to fix it, we have to change `headers.put(Header::CONTENT_TYPE, m_contentType);` to `headers.putIfNotExists(Header::CONTENT_TYPE, m_contentType);`
> here - https://github.com/oatpp/oatpp/blob/master/src/oatpp/web/protocol/http/outgoing/BufferBody.cpp#L64

*oops*
I should have read the issue carefully 


> What if we rename `replaceOrPut` --> `putOrReplace` - this way all `put` related methods would start with `put` word

@lganzzzo Yes, I thought so myself but then came to the conclusion, If I would want to replace something, I would first search something beginning with replace. Not every IDE is as smart as CLion which would find it anyways.
I have no strong opinion on the particular wording. If you think all beginning with `put` would be more consistent, I'll happily change the names and would merge the PR.

lganzzzo:
@bamkrs ,

> If you think all beginning with put would be more consistent, I'll happily change the names and would merge the PR.

Yes, let's do it.
Please go ahead and merge the PR after those changes.

lganzzzo (approved):

PR #484, opened 2021-09-15
Add setter for `Server`s `ConnectionProvider`.

bamkrs:
We may need synchronization for that.

### DIFF
--- a/src/oatpp/network/Server.cpp
+++ b/src/oatpp/network/Server.cpp
@@ -50,11 +50,13 @@ void Server::conditionalMainLoop() {
 
   while (getStatus() == STATUS_RUNNING) {
     if (m_condition()) {
+      m_spinlock.lock();
       auto connection = m_connectionProvider->get();
-
+      m_spinlock.unlock();
       if (connection) {
         if (getStatus() == STATUS_RUNNING) {
           if (m_condition()) {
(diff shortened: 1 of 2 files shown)


src/oatpp/network/Server.cpp
```diff
@@ -163,6 +163,14 @@ v_int32 Server::getStatus() {
   return m_status.load();
 }
 
+void Server::setConnectionProvider(const std::shared_ptr<ServerConnectionProvider> &connectionProvider) {
+  m_connectionProvider = connectionProvider;
```
lganzzzo, line 167: I think we should protect with the spin-lock.
Since server run loop is running in a different thread.

bamkrs: Yes, just wanted to hear your opinion.

src/oatpp/network/Server.cpp
```diff
@@ -50,11 +50,13 @@ void Server::conditionalMainLoop() {
 
   while (getStatus() == STATUS_RUNNING) {
     if (m_condition()) {
+      m_spinlock.lock();
       auto connection = m_connectionProvider->get();
```
lganzzzo, line 54: `ConnectionProvider::get()` can potentially throw.
It's better to use lock guard.

lganzzzo (approved):

PR #486, opened 2021-09-27
Add ErrorHandler virtual destructor

Garzet:
Virtual destructor added to the ErrorHandler since derived class
destructor should be called when destroying an object via a pointer to
the base type (ErrorHandler in this case).

More info: https://stackoverflow.com/a/461224/6067014


lganzzzo:
Nice catch!
Merged!

PR #492, opened 2021-10-14
Add option to include required fields in JSON even if `value == nullptr` and `includeNullFields == false`.

bamkrs:
In some situations, some fields without a value still need to be included in the serialized JSON, even if `includeNullFields == false`.
With this PR, one can set the serializer-config's `honorRequired` to `true` and the serializer will include the null-field even if `includeNullFields` is set to false.

### DIFF
--- a/src/oatpp/parser/json/mapping/Serializer.cpp
+++ b/src/oatpp/parser/json/mapping/Serializer.cpp
@@ -160,7 +160,7 @@ void Serializer::serializeObject(Serializer* serializer,
   for (auto const& field : fields) {
 
     auto value = field->get(object);
-    if (value || config->includeNullFields || (field->info.required && config->honorRequired)) {
+    if (value || config->includeNullFields || (field->info.required && config->alwaysIncludeRequired)) {
       (first) ? first = false : stream->writeSimple(",", 1);
       serializeString(stream, (p_char8)field->name, std::strlen(field->name));
       stream->writeSimple(":", 1);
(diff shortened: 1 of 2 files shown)


bamkrs:
Hey @lganzzzo , will do!

I have another question. Would it be possible to archive something like this:

```c++
class StatusDto : public oatpp::DTO {
  DTO_INIT(StatusDto, DTO)
  DTO_FIELD(String, status);
  DTO_FIELD(Int32, code);
  DTO_FIELD(String, message);
};
```

Now with an serializer set to `includeNullFields = false`, the following snipped would yield
```c++
dto->status = "Ok";
dto->code = 200;
dto->message = "Okay";
// -> {"status":"Ok","code":200,"message":"Okay"}
```
```c++
dto->status = "Ok";
dto->code = 200;
// -> {"status":"Ok","code":200}
```
Right? Right.
But would it be possible to archive something like:
```c++
dto->status = "Ok";
dto->code = 200;
dto->message = oatpp::Void();
// -> {"status":"Ok","code":200, "message":null} 
```
Even with `includeNullFields = false` without fumbling with the `alwaysIncludeRequired`?

There is a special use-case with an client-side library we are using and we need to address all 3 outputs on an per-object-basis in a list.

bamkrs:
> Also, there is a third variant, closer to what you proposed:
> 
> * change message type to `oatpp::Any`
> * add your custom type ex.: `oatpp::Null`
> * add serializer method for `oatpp::Null`
> * assign string to message if you need message to be a string, leave message null if you dont want to show it, assign `oatpp::Null` to message if you want it to be an explicit `null`

Aaah yes, this is exactly what I thought of. Using the flexibility of custom types is a great solution for that!

Issue #483, opened 2021-09-13
SSL Certificate reread on HUP signal


acidtonic:
On Unix it would be nice if oatpp could reread the configuration files related to SSL certificates such that they could be updated without restarting the server.

Normally what would happen is the following...

1. SSL files get updated.
2. Oatpp server receives HUP signal
3. Oatpp rereads the certificate without refusing any connections or dropping existing connections still using the previous certificate.

Although we should have an HA environment in production, In my opinion a failover shouldn't be needed to change certificates and remain operational.

bamkrs:
Hey @acidtonic 
That is indeed a very good point.
However, this will be a tough one. The main issue is in Oat++'s component system. I see two options here. Introduce "Factory-Components" that run their lambda every time the component is requested or implement the `XYZssl` components as some kind of stateful factory which creates new library-instances on HUP signals. 

However, the easiest, but not cleanest way, would be to introduce a new function to the `XYZssl`'s `ConnectionProvider` which you can pass a new configuration. This function would then handle the creation of new library instances etc.

bamkrs:
I created an PR to show my last proposed approach, its a bit overkill since there is an more easy way.

As pointed by @lganzzzo in https://github.com/oatpp/oatpp-openssl/pull/8#issuecomment-919750055 it's way easier to just create an entire new `ConnectionProvider` with a new setter-function for the `ConnectionProvider` in `Server`.
This will need some manual labor since you cant just use the components anymore and have to hand-build them.

bamkrs:
With https://github.com/oatpp/oatpp/pull/484 the workflow would be something like that:

```c++
std::shared_ptr<oatpp::network::Server> server;

std::shared_ptr<oatpp::network::server::ConnectionProvider> createNewProvider() {
    auto config = oatpp::libressl::Config::createDefaultServerConfigShared(CERT_CRT_PATH, CERT_PEM_PATH /* private key */);
    return oatpp::libressl::server::ConnectionProvider::createShared(config, {"0.0.0.0", 8443, oatpp::network::Address::IP_4});
}

void signal_handler(int sig) {
  if (sig == SIG_HUP) {
    server->setConnectionProvider(createNewProvider());
  }
}

void you_run_method() {
  // components, controllers etc
  // [...]

  /* create server */
  server = oatpp::network::Server::createShared(createNewProvider(), OATPP_GET_COMPONENT(std::shared_ptr<ConnectionHandler>));
  server.run()

}
  
```

[...]

Issue #476, opened 2021-09-05
Server shutdown while connection is still open


heindl-solutions:
We have set up the server similar to
https://github.com/oatpp/example-server-stop/blob/master/src/App_StopByConditionCheck.cpp

If a client opens a connection to `localhost:8000`, the function `HttpProcessor::Task::run()` is entered 3 times, and left immediately 2 times.
This means that one thread handling the open connection is kept open, as might be expected.

But now there is a problem if we want to shut down the server, because the loop (HttpProcessor.cpp, l. 237)
```
 do {
      connectionState = HttpProcessor::processNextRequest(resources);
    } while (connectionState == ConnectionState::ALIVE);
```

is still running until the client closes the connection, although we have stopped and deleted the underlying resources already!

Thus it seems there is currently no way to shut down the server in a clean way.

We are using Oatpp 1.2.5 on x64-win64.

bamkrs:
Hi Andreas, 
yes, currently the stopping of the server is only defined for new connections. This way, already established connections can be handled and gracefully closed, thus preventing dangling I.E. Database calls or file writes (if correctly implemented/synchronised on the user side). Oat++ was never designed to be shut down in the first place so all those functions are gradually added. 

As of version 1.3.0 (upcoming release) for example, we at least count the created connections and keep everything alive until the last connection was served (`stop()` actually waits/blocks for connections to be closed!). Thus, we track everything inside Oat++ and the user does not need to anymore. 
Maybe give the 1.3.0 branch a try and see if it fixes your problem.

sgallou:
I encoutered the same problem, which makes my server crash (under Windows) after processing a request containing a `Connection=keep-alive` header. 

The workaround I found is to use my own connection provider, which is keeping trace of created connections, to be able to close they (= free sockets) :

```cpp
class MyConnectionProvider final : public oatpp::network::tcp::server::ConnectionProvider
{
public:
   MyConnectionProvider(const oatpp::network::Address& address,
      bool useExtendedConnections = false)
      : ConnectionProvider(address,
         useExtendedConnections)
   {
   }

   ~MyConnectionProvider() override
   {
      for (const auto& c : m_connections)
         c->close();
   }

   static std::shared_ptr<MyConnectionProvider> createShared(const oatpp::network::Address& address,
      bool useExtendedConnections = false)
   {
      return std::make_shared<MyConnectionProvider>(address,
         useExtendedConnections);
   }

[...]

PR #495, opened 2021-10-20
Better provider interface

lganzzzo:
- Provider pattern improved. Resource Provider is decoupled to Provider and Invalidator
- Introduced ConnectionProviderSwitch - for proper hot reloads (#483)
- ConnectionHandlers are now invalidating all connections on stop. Which makes it possible to stop the server correctly. (#476)


bamkrs:
d74d6b2308bc9f7819d545481a2b1925277b00f6 is a huge step forward for so many things. Nice!

PR #501, opened 2021-10-29
Fix errors

NickPak:


src/oatpp/network/tcp/client/ConnectionProvider.cpp
```diff
@@ -309,7 +309,7 @@ oatpp::async::CoroutineStarterForResult<const provider::ResourceHandle<data::str
 
 #endif
 
-      m_currentResult = m_currentResult->ai_next;
+      // m_currentResult = m_currentResult->ai_next;
```
lganzzzo, line 312: Hello @NickPak ,
Why have you commented this out?

PR #505, opened 2021-11-01
Fix errors

NickPak:


src/oatpp/network/tcp/client/ConnectionProvider.cpp
```diff
@@ -289,7 +289,7 @@ oatpp::async::CoroutineStarterForResult<const provider::ResourceHandle<data::str
       }
       if(error == WSAEWOULDBLOCK || error == WSAEINPROGRESS) {
         return ioWait(m_clientHandle, oatpp::async::Action::IOEventType::IO_EVENT_WRITE);
-      } else if(error == WSAEINTR) {
+      } else if(error == WSAEINTR || error == WSAEALREADY) {
```
lganzzzo, line 292: This looks like a possible valid fix.
Also, it looks like we have to also check for `WSAEINVAL` too - see https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-connect

lganzzzo:
Merged!
Great work!

PR #508, opened 2021-11-05
fix: wrong ANSI color codes in DefaultLogger::log()

aeschbacher:
This patch restores correct colorized output for `OATPP_LOGx()`

Tested on Linux only

### DIFF
--- a/src/oatpp/core/base/Environment.cpp
+++ b/src/oatpp/core/base/Environment.cpp
@@ -76,23 +76,23 @@ void DefaultLogger::log(v_uint32 priority, const std::string& tag, const std::st
 
   switch (priority) {
     case PRIORITY_V:
-      std::cout << "\033[0;0m V \033[0m|";
+      std::cout << "\033[0m V \033[0m|";
       break;
 
     case PRIORITY_D:
-      std::cout << "\033[34;0m D \033[0m|";
+      std::cout << "\033[34m D \033[0m|";
       break;
 
     case PRIORITY_I:
-      std::cout << "\033[32;0m I \033[0m|";
+      std::cout << "\033[32m I \033[0m|";
       break;
 
     case PRIORITY_W:
-      std::cout << "\033[45;0m W \033[0m|";
+      std::cout << "\033[45m W \033[0m|";
       break;
 
     case PRIORITY_E:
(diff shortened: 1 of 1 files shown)


src/oatpp/core/base/Environment.cpp
```diff
@@ -76,23 +76,23 @@ void DefaultLogger::log(v_uint32 priority, const std::string& tag, const std::st
 
   switch (priority) {
     case PRIORITY_V:
-      std::cout << "\033[0;0m V \033[0m|";
+      std::cout << "\033[0m V \033[0m|";
```
lganzzzo, line 79: I'm not very familiar with color codes,

But they seem to differ between Mac, Linux, Windows.

Wich OS this fix for?
Should we add any ifdefs here?

aeschbacher: I think that the ability to render [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) is not OS specific, but depends on the terminal emulator used. 
- On Unix-like systems, most (all?) terminal emulators support ANSI colors
- On MacOS, Terninal.App and iTerm2 support ANSI colors.
- On Windows, it's more complicated:
  - it works fine with WSL
  - for CMD / PowerShell, it _seems_ that it should be possible activate ANSI colors by [adding a key in the registry](https://www.codeproject.com/Tips/5255355/How-to-Put-Color-on-Windows-Console
)
  ```
  [HKEY_CURRENT_USER\Console]
  "VirtualTerminalLevel"=dword:00000001
  ```
  but I cannot test it.


Note: there are other places in oatpp's code where ANSI colors are correct:
- https://github.com/oatpp/oatpp/blob/master/src/oatpp-test/UnitTest.cpp#L35
- https://github.com/oatpp/oatpp/blob/master/src/oatpp-test/UnitTest.cpp#L56
- https://github.com/oatpp/oatpp/blob/master/src/oatpp/core/base/Environment.hpp#L494

It could indeed be interesting to introduce a cmake option to enable/disable ANSI colors (not really sure it's worth it though)

lganzzzo: Hey, thanks for the detailed description!

lganzzzo (approved):

lganzzzo:
Merged!

Issue #520, opened 2021-11-25
CPack for DEB and RPM is missing from the cmake files


julian-st:

lganzzzo:
Fixed in #524
Thanks to @bf-bryants

PR #524, opened 2021-12-01
Feature/cpack

bf-bryants:
This provides a fix for https://github.com/oatpp/oatpp/issues/520 and adds some CPack functionality to CMake.  After building, call `cpack -G DEB` or `cpack -G RPM` in the build directory to generate installable packages.

The configuration sets up two packages in classical form: an oatpp "runtime" for the shared library, and a developer package for compiling against oatpp for the headers and generated CMake package.  The "runtime" package is only created when oatpp is built as a shared library.  CMake appears to default to building the library as static; this can be changed by calling CMake with `-DBUILD_SHARED_LIBS=ON`.

Tested on Ubuntu (DEB) and OpenSUSE (RPM).  The NSIS configuration might work on Windows, but this was not tested.


lganzzzo:
Thanks for the great PR!
Merged!

PR #527, opened 2021-12-03
fix: add port in host header

### LINKED ISSUE #525: Host header does not contain port
Currently the "Host" header added to any HTTP request only contains the `host` parameter and ignores the `port` parameter.

https://github.com/oatpp/oatpp/blob/4a40edc0c38feb04d9b3d73bde02ef91e5878fb2/src/oatpp/web/client/HttpRequestExecutor.cpp#L201

The [HTTP/1.1 specification](https://httpwg.org/specs/rfc7230.html#header.host) implies that if the port is not defined, the default HTTP or HTTPS port should be used.

I'm sending a request on a service running on port 8080 and it refuses my request as the Host header does not match.

Is this behavior wanted on your side for any reason or can I make a PR to fix this?

Thanks!

ArnaudGallardo:
Closes #525 

I used a `BufferOutputStream` to concatenate strings, don't hesitate to tell me if there is a better way to do it!

### DIFF
--- a/src/oatpp/web/client/HttpRequestExecutor.cpp
+++ b/src/oatpp/web/client/HttpRequestExecutor.cpp
@@ -29,6 +29,7 @@
 
 #include "oatpp/network/tcp/Connection.hpp"
 
+#include "oatpp/core/data/stream/BufferStream.hpp"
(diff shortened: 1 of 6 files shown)


lganzzzo:
Looks good!
Merged!

PR #528, opened 2021-12-05
handle different decimal separators in floating point JSON serialization

### LINKED ISSUE #201: Decimal comma in Float32 in JSON
If I set the locale of my Ubuntu System to "de_DE.UTF-8" oat++ serializes float numbers with a comma as decimal separator instead of a dot.
When executing a GET using the Swagger UI I get:

can't parse JSON.  Raw result:
{"x":2,000000,"y":0,000000,"z":200,000000}

Is this legal JSON or a bug in oat++ or can this be configured somehow?

langchr86:
Fixes #201 

Based on the discussion in https://github.com/oatpp/oatpp/issues/201 I tried to find a simple but in most of the time not (significant) slower solution for the JSON floating point number serialization/parsing. 

Problem:
A floating point number in JSON uses always the "." character as decimal separator (radix) as defined here: https://www.rfc-editor.org/rfc/rfc7159.html#section-6. Because oatpp uses `std::strtof` and `std::strtod` for float parsing and `snprintf` for serialization the implementation does not always use ".". Instead these functions respect the locale set to the process. Therefore in some locales a different character as "." are used. E.g. in "de_*" locations a "," is used. See: https://en.wikipedia.org/wiki/Decimal_separator or https://docs.oracle.com/cd/E19455-01/806-0169/overview-8/index.html.

Solution:
My proposal is based on the check if we are in a locale where the separator is different from the needed ".". This is done by using https://en.cppreference.com/w/c/locale/localeconv to query the currently used separator. If it matches "." nothing is changed in the current implementations.
But if the character is different we try to prepare the to serialize/parse string to match the expected separator character by `std::strtof`/`std::strtod`/`snprintf`. Therefore we replace the locale defined separator with "." after serialization and the received "." by the expected locale defined separator before parsing.

Notes:
This is just a proposal to check if it is worth to continue working on this. Further useful work would consist of:
* Implement functional unit tests.
* Implement and document benchmark.
* Refactor code. E.g. to have the `json_decimal_point` only defined in one central point.


src/oatpp/core/utils/ConversionUtils.cpp
```diff
@@ -189,11 +190,30 @@ namespace oatpp { namespace utils { namespace conversion {
   }
 
   v_buff_size float32ToCharSequence(v_float32 value, p_char8 data, v_buff_size n, const char* format) {
-    return snprintf((char*)data, n, format, value);
+    return float64ToCharSequence(value, data, n, format);
```
lganzzzo, line 193: Please do not change the behavior of this method.
Instead, please add a new method to `json::Utils` 
https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/Utils.hpp

src/oatpp/core/utils/ConversionUtils.cpp
```diff
@@ -189,11 +190,30 @@ namespace oatpp { namespace utils { namespace conversion {
...
   }
 
-v_buff_size float64ToCharSequence(v_float64 value, p_char8 data, v_buff_size n, const char* format) {
-    return snprintf((char*)data, n, format, value);
+  v_buff_size float64ToCharSequence(v_float64 value, p_char8 data, v_buff_size n, const char* format) {
```
lganzzzo, line 196: Please do not change the behavior of this method.
Instead, please add a new method to json::Utils
https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/Utils.hpp

src/oatpp/core/utils/ConversionUtils.cpp
```diff
@@ -189,11 +190,30 @@ namespace oatpp { namespace utils { namespace conversion {
...
+    const char locale_decimal_point = localeconv()->decimal_point[0];
+    static constexpr char json_decimal_point = '.';
+    if (locale_decimal_point != json_decimal_point) {
+      convertSingleDecimalPointInCharSequence(data, locale_decimal_point, json_decimal_point);
+    }
+
+    return ret;
+  }
+
+  p_char8 convertSingleDecimalPointInCharSequence(p_char8 data, char old_char, char new_char) {
```
lganzzzo, line 208: Please move this method to `json::Utils` - 
https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/Utils.hpp

lganzzzo (changes requested):

lganzzzo:
Hello @langchr86 ,

Thanks for PR fixing a long-time bug!

The proposed solution looks good.
The only comment - please move utility methods to [json::Utils](https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/Utils.hpp) - since those are json specifics.

Regards,
Leonid

langchr86:
Thank you for your feedback.
I restructured the added code to have all new functions in the proposed `json::Utils` class.
I also added unit tests for these new functions. They are only executed in a system where the needed locales are available. Maybe we could update the CI environment to meet these requirements or check which matching locales are already available and change the tests accordingly.
Originally I wanted to execute some benchmarks. Do you already have some infrastructure which can be used? Or should I simply write throw-away code for e.g. google-benchmark and paste only the results here? What is your suggestion?

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -399,7 +400,7 @@ v_io_size ConsistentOutputStream::writeAsString(v_uint64 value){
 
 v_io_size ConsistentOutputStream::writeAsString(v_float32 value){
   v_char8 a[100];
-  auto size = utils::conversion::float32ToCharSequence(value, &a[0], 100);
+  auto size = parser::json::Utils::float32ToJson(value, &a[0], 100);
```
lganzzzo, line 403: Since `ConsistentOutputStream` is a general-purpose class - it should
use `auto size = utils::conversion::float32ToCharSequence(value, &a[0], 100);` as before

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -408,7 +409,7 @@ v_io_size ConsistentOutputStream::writeAsString(v_float32 value){
 
 v_io_size ConsistentOutputStream::writeAsString(v_float64 value){
   v_char8 a[100];
-  auto size = utils::conversion::float64ToCharSequence(value, &a[0], 100);
+  auto size = parser::json::Utils::float64ToJson(value, &a[0], 100);
```
lganzzzo, line 412: Since `ConsistentOutputStream` is a general-purpose class - it should
use `auto size = utils::conversion::float64ToCharSequence(value, &a[0], 100);` as before

src/oatpp/core/parser/Caret.cpp
```diff
@@ -375,6 +377,9 @@ v_int64 Caret::StateSaveGuard::getSavedErrorCode() {
   v_float32 Caret::parseFloat32(){
     char* end;
     char* start = (char*)&m_data[m_pos];
+
+    parser::json::Utils::convertSingleDecimalSeparatorFromJsonToLocale((p_char8) start);
```
lganzzzo, line 381: We can't add this here - since `Caret` is a general-purpose class.
It's not serving JSON only.

src/oatpp/core/parser/Caret.cpp
```diff
@@ -386,6 +391,9 @@ v_int64 Caret::StateSaveGuard::getSavedErrorCode() {
   v_float64 Caret::parseFloat64(){
     char* end;
     char* start = (char*)&m_data[m_pos];
+
+    parser::json::Utils::convertSingleDecimalSeparatorFromJsonToLocale((p_char8) start);
```
lganzzzo, line 395: We can't add this here - since `Caret` is a general-purpose class.
It's not serving JSON only.

lganzzzo (changes requested):

src/oatpp/core/parser/Caret.cpp
langchr86: Could you explain where the correct entry points for JSON serialization/parsing are?

lganzzzo: Hey @langchr86 ,

My bad, sorry for not providing this information earlier.

So this PR is specific to JSON, so we want to keep this for JSON only. We don't want to alter the behavior of general-purpose methods like `utils::conversion::float64ToCharSequence`.

Here is where float gets serialized in JSON - https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/mapping/Serializer.hpp#L124

And here where float gets deserialized from JSON - 
https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/mapping/Deserializer.cpp#L157

My proposition is to put the fix right where it gets serialized/deserialized and if you need some helper methods - then you can use `json::Utils` class to put those there. We can use specialized templates or whatever is needed. You may also probably need a helper method to parse float from the `Caret` - you can also put that method to `json::Utils`.

Please let me know if you need more info or any help.

Regards,
Leonid

langchr86: Thank you for your input. I refactored the whole implementation to meet your suggestions. I also added more tests for the new (and old) features of parsing/serializing.

One thing that is kind of an ugly design is that in the `Deserializer::deserializeFloatXX` methods the constness of the pointer returned by `Caret::getCurrData()` is removed to be able to replace possible different decimal separators in the internally hold string data. But I see no easy/cheap way to improve this. Because the whole parsing is designed around the non-mutable `Caret` class. And an approach where we first copy the part of the string that contains the float number is expensive and complicated. Because first we would kind of parse the length of the float number (this would kind of involve "real" float parsing), then copy it and then do the replacement of decimal separators and parsing. Instead of something like that we implement the whole parsing of JSON by ourself. Or we would need to copy the whole data of `Caret` which could be a big JSON structure.

At the end my current implementation is pragmatic and should meet your requirements. I hope this is OK for your.

lganzzzo (changes requested):
Please make sure that the original text isn't modified during the parsing.

lganzzzo:
> Originally I wanted to execute some benchmarks. Do you already have some infrastructure which can be used? Or should I simply write throw-away code for e.g. google-benchmark and paste only the results here? What is your suggestion?

I have some recommendations for performance improvements - but first, let's have the correct implementation for this improvement.

Also, it's understood that this improvement will affect performance a bit - I'm ok with that.

src/oatpp/core/parser/Caret.cpp
```diff
@@ -180,6 +180,10 @@ v_int64 Caret::StateSaveGuard::getSavedErrorCode() {
   v_buff_size Caret::getPosition(){
     return m_pos;
   }
+
+  void Caret::advancePosition(v_buff_size n){
```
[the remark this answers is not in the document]
langchr86: Sorry for that. Remove it.

src/oatpp/parser/json/mapping/Deserializer.cpp
```diff
@@ -161,9 +161,22 @@ oatpp::Void Deserializer::deserializeFloat32(Deserializer* deserializer, parser:
...
   if(caret.isAtText("null", true)){
     return oatpp::Void(Float32::Class::getType());
-  } else {
-    return Float32(caret.parseFloat32());
   }
+
+  const char* start = caret.getCurrData();
+  const char* end = caret.getData() + caret.getDataSize();
+  // here we remove constness and manipulate the internal memory of Caret directly
+  parser::json::Utils::convertFirstDecimalSeparatorFromJsonToLocale((p_char8) start, (p_char8) end);
```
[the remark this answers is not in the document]
langchr86: Ok. I now added another commit with a preview of another approach. I will clean up later.
This new approach combines the decimal separator replacing and the string extraction into a separate buffer. Like this we save another loop after the extraction for the decimal separator replacement. Please let me know your thoughts, that I can cleanup and finish later.

lganzzzo (commented):
Hey @langchr86 ,

The latest changes look good to me.

I have some minor things I want to improve - 
I've forked your repo and later will create a PR to your branch for you to review.

rozmansi:
Any news on this one? Was able to set `setlocale(LC_NUMERIC, "C")` in my project for the time being, but hoping for a proper fix. Pretty please?

PR #529, opened 2021-12-13
allow oatpp::Any to parse directly to UInt64 or Int64

langchr86:
Currently the parsing of `oatpp::Any` can only parse numbers to `oatpp::Float64`.
In my use case I need exact parsing for integer types. Therefore we need to guess the integer type of a JSON "integer" value.

This PR extends the `Deserializer::guessNumberType` function to search for a decimal separator. If no is found in the next word we have an integer value. Depending on the first character it is then parsed as `UInt64` or `Int64`.

I also added tests that checks all supported cases for `oatpp::Any` parsing.

### DIFF
--- a/src/oatpp/parser/json/Utils.cpp
+++ b/src/oatpp/parser/json/Utils.cpp
@@ -476,6 +476,21 @@ std::string Utils::parseStringToStdString(ParsingCaret& caret){
   return "";
   
 }
-    
+
+bool Utils::findDecimalSeparatorInCurrentNumber(ParsingCaret& caret) {
+  parser::Caret::StateSaveGuard stateGuard(caret);
+
+  // search until a decimal separator is found or no more digits are found or no more data available
+  while(caret.canContinue()) {
+    if (caret.isAtChar(JSON_DECIMAL_SEPARATOR)) {
+      return true;
+    }
+    if (!caret.isAtDigitChar()) {
+      return false;
+    }
+    caret.inc();
+  }
+  return false;
+}
   
 }}}
--- a/src/oatpp/parser/json/Utils.hpp
+++ b/src/oatpp/parser/json/Utils.hpp
@@ -61,6 +61,11 @@ public:
    */
   static constexpr v_int64 ERROR_CODE_PARSER_QUOTE_EXPECTED = 3;
 
+  /**
+   * Character that is used as decimal separator in floating point string representations.
+   */
+  static constexpr v_char8 JSON_DECIMAL_SEPARATOR = '.';
+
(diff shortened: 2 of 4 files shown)


langchr86:
Looks like the problems with CI is not related to my code changes but with the CI infrastructure.

src/oatpp/parser/json/Utils.cpp
```diff
@@ -476,6 +476,15 @@ std::string Utils::parseStringToStdString(ParsingCaret& caret){
   return "";
   
 }
-    
+
+bool Utils::findDecimalSeparatorInNextWord(const char* data) {
```
lganzzzo, line 480: This method take `ParsingCaret& caret` as a parameter because we are bounded by caret.dataSize and not by '\0' char.

src/oatpp/parser/json/Utils.cpp
```diff
@@ -476,6 +476,15 @@ std::string Utils::parseStringToStdString(ParsingCaret& caret){
...
+  while (*data != '\0' && std::isspace(*data) == 0) {
```
lganzzzo, line 481: This condition should be changed. 
It checks for `\0` char or until `std::isspace == false` which is incorrect.
Consider the following JSON:

```json
{"a":1,"b":1.1}
```

in this case this method will be evaluated to true when checking `a` since it won't stop on the `,`.

lganzzzo (changes requested):
Thanks for the Great PR!

Please fix the `findDecimalSeparatorInNextWord()` method

src/oatpp/parser/json/Utils.cpp
langchr86: Good point. Is there any helper or definition which characters can mark the end of a "value field"? Or should I introduce something globally for the "," and "}"? Or simply use these locally in this method?

lganzzzo: I think we can do the following check - if the current char is anything but a digit or a decimal separator - stop.

> Or should I introduce something globally for the "," and "}"? Or simply use these locally in this method?

Locally should be enough

langchr86: Thank you for your inputs. I changed the implementation accordingly and tried to use `Caret`. I hope I did understand the usage of `Caret` and `StateSaveGuard` correctly. I also improved the unit tests for numbers a bit to have the case tested where additional content comes after the parsed number.

lganzzzo (approved):

lganzzzo:
Thanks for the great PR!
Merged!

PR #541, opened 2022-01-06
Only build oatpp-test if OATPP_BUILD_TESTS is YES

dencrypt-jacob:
I noticed that even when building with OATPP_BUILD_TESTS = NO, an oatpp-test library would still be built and other oatpp modules would still depend on it.

So these modifications remove all oatpp-test, when  OATPP_BUILD_TESTS = NO.

I will have similar pull requests for oatpp-mbedtls and oatpp-websocket.

I am not using the remaining oatpp modules, so I have not checked how this would affect them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,9 +9,6 @@ string(REGEX REPLACE "#define OATPP_VERSION \"([0-9]+.[0-9]+.[0-9]+)\"$" "\\1" o
 
 set(OATPP_THIS_MODULE_NAME oatpp) ## name of the module (also name of folders in installation dirs)
 set(OATPP_THIS_MODULE_VERSION ${oatpp_VERSION}) ## version of the module (also sufix of folders in installation dirs)
-set(OATPP_THIS_MODULE_LIBRARIES oatpp oatpp-test) ## list of libraries to find when find_package is called
-set(OATPP_THIS_MODULE_TARGETS oatpp oatpp-test) ## list of targets to install
-set(OATPP_THIS_MODULE_DIRECTORIES oatpp oatpp-test) ## list of directories to install
 
 ###################################################################################################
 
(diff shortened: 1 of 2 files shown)


lganzzzo:
Hello @dencrypt-jacob ,

`oatpp-test` is a small library for testing oatpp applications and it's always included with oatpp.
`OATPP_BUILD_TESTS` - is a flag that signals whether or not build tests for oatpp (liboatpp itself)

Basically, you may not be interested in testing oatpp but you still want to test your oatpp-based application.

lganzzzo:
This PR won't be merged.

Please feel free to text/comment in case you have further questions/ideas

Best Regards,
Leonid

dencrypt-jacob:
Hmmm - OK. The problem is that I build a "core" for use in iOS and Android apps. I provide shared libraries for oatpp, oatpp-mbedtls and oatpp-websocket + a lot of other stuff. Now I will have to provide oatpp-test as well, even though, I am pretty sure the clients will not be using it.
Maybe the issue should be solved differently. It seems oatpp is not depending on oatpp-test, but oatpp-mbedtls is depending on it, so the linker requires oatpp-test to be present if I try to load oatpp-mbedtls.

What do you think?

dencrypt-jacob:
With you comments in mind, I will take a closer look, if I can avoid the dependency in oatpp-mbedtls.

lganzzzo:
`oatpp`, `oatpp-mbedtls`, `oatpp-websocket` - can be built without `oatpp-test` library.
`oatpp-test` is used only for test-suites.

In order to build modules without `oatpp-test` though you have to change CMake scripts.
It's just the way we currently have this organized.

We had similar inquiries from other guys/orgs but as far as I know everyone ended up just ignoring the presence of `oatpp-test` and linking it as it works now with current build scripts (since oatpp-test is a very small lib).

lganzzzo:
However, we can think of some solution - like an additional option in all oatpp modules - whether or not link `oatpp-test`.
Please file an issue for that

dencrypt-jacob:
OK. The changes to the CMake scripts are fairly small. I have already made them. I'll make a PR for oatpp-mbedtls and you can comment on that. If you don't want it, no problem, then I'll also live with the small extra library.
But ideally it would be nice to be able to compile without the dependency.

Thanks for the feedback.

lganzzzo:
Ok. I've reviewed this PR once again
Here is what we have to change:
- `OATPP_BUILD_TESTS` flag should do the same as before - if ON - we build tests for oatpp if OFF we don't build tests for oatpp.
- Introduce another flag something like ex.: `OATPP_LINK_TEST_LIBRARY` or `OATPP_LINK_OATPP_TEST` default `ON` - when `ON` - we link `oatpp-test` when `OFF` we don't. The same flag with the exact same name should be introduced in other modules.

I think in this case we can go ahead with this PR.

dencrypt-jacob:
OK. I think these flags are tightly connected, and there are some places, where I am in doubt wether to use one or the other, but I think it is easier to discuss from a PR - then we can iterate towards a solution.
I'll do an update and you can let me know, if this is how you want it.

PR #544, opened 2022-01-11
optimize includes

julianreisch:
We applied our tool for optimizing includes to the oatpp project. It ensures for each file that header files are included if and only if they are used. As can be seen, many includes can be moved from header to cpp files. This reduces the number of lines of code after the preprocessor (in .o files) by 6.2% which results in a shorter compilation time. We computed only 190.3 sec user time on average on our machine, compared to 217.4 sec without the optimization - a reduction by 12.45%.


lganzzzo:
This is actually an interesting contribution. I can see there were troubles with Windows and Mac builds...

julianreisch:
That sounds promising :) We really could use your help in this regard.
Also, I made a mistake and many more includes can be removed (so that there are not 1,260 insertion and 431 deletions, but 144 insertions and 464 deletions). 
I will just reopen the PR then.

src/oatpp/core/concurrency/Thread.cpp
```diff
@@ -23,9 +23,10 @@
  ***************************************************************************/
 
 #include "Thread.hpp"
+#include "oatpp/core/base/Environment.hpp"
 
 #if defined(_GNU_SOURCE)
-  #include <pthread.h>
+#include <thread>
```
lganzzzo, line 29: I think this guy should be  `<pthread.h>` since we are using `pthread_setaffinity_np` method

lganzzzo (commented):
Looks great!
Most probably we'll have to merge it to `1.4.0` (I need to create a branch first)
Since includes shuffling most probably will cause CI failures in user projects.

lganzzzo:
Please change target branch to [v1.4.0](https://github.com/oatpp/oatpp/tree/v1.4.0)

lganzzzo:
@julianreisch , FYI, I've changed the target branch to `1.4.0`

lganzzzo (approved):

bamkrs:
This is huge! Looking forward for 1.4 already!

lganzzzo:
Merged!
Great Work!

Issue #557, opened 2022-02-03
ClassId are euqal, but Type::extends() return false


yikenan:
Hello, I have a question about oatpp::Any.
My test code is as follows，see the comments for my questions, thank you.

```
#include OATPP_CODEGEN_BEGIN( DTO )

class DtoTest : public oatpp::DTO
{
    DTO_INIT( DtoTest, DTO )
    DTO_FIELD( Any, test );
};

#include OATPP_CODEGEN_END( DTO )

const char *szJson = "{\"test\" : 123}";

int main()
{
    oatpp::parser::json::mapping::ObjectMapper objMapper;
    auto spDto = objMapper.readFromString<oatpp::Object<DtoTest>>( szJson );

    auto pClassType = spDto->test.getStoredType();
    if ( pClassType->classId == oatpp::Float64::Class::getType()->classId )
    {
        std::cout << "classId are equal, classId.id = " << pClassType->classId.id
                  << ", classId.name = " << pClassType->classId.name << std::endl;
    }

    if ( !pClassType->extends( oatpp::Float64::Class::getType()))
    {
        // My expectation is true. Should it be true? Is it a bug?
        // Therefore, the retrieve() function cannot be used and an exception will be thrown
        // spDto->test.retrieve<oatpp::Float64>();
        std::cout << "extends() return false" << std::endl;
    }

    if ( pClassType != oatpp::Float64::Class::getType())
    {
        // extends() return false because the pointers are not equal
        // Shouldn't they be equal?
        std::cout << "Pointers are not equal" << std::endl;
    }

    return 0;
}
```

Run the code, output is as follows

```
classId are equal, classId.id = 15, classId.name = Float64
extends() return false
Pointers are not equal

```

yikenan:
My environment:
Oatpp 1.3.0
gcc 9.3.0

yikenan:
I ran the oatpp test program oatpp alltests, which has the same code, but there is no problem:

```
 I |2022-02-03 18:37:55 1643884675074373| TEST[parser::json::mapping::DeserializerTest]:START...
 D |2022-02-03 18:37:55 1643884675074458| TEST[parser::json::mapping::DeserializerTest]:Any: String
 D |2022-02-03 18:37:55 1643884675074484| TEST[parser::json::mapping::DeserializerTest]:Any: Boolean
 D |2022-02-03 18:37:55 1643884675074500| TEST[parser::json::mapping::DeserializerTest]:Any: Negative Float
[...]

PR #558, opened 2022-02-04
move the getType() function implementation from hpp file to cpp file

yikenan:
see https://github.com/oatpp/oatpp/issues/557


lganzzzo:
Merged!
Thanks for the PR!

PR #565, opened 2022-02-20
Add new virtual handleError method to handle the original error

### LINKED ISSUE #277: API question - ErrorHandler
In file: oatpp/web/server/handler/ErrorHandler.hpp

About the method:
```cpp
namespace oatpp { namespace web { namespace server { namespace handler {
class ErrorHandler {
// [...]
  std::shared_ptr<protocol::http::outgoing::Response> handleError(const protocol::http::Status& status, const oatpp::String& message);
};
```
# Question

While implementing an ErrorHandler I noticed the error is not provided to `handleError` but only it's message `error->what()`
as defined in:
```cpp
class ApiController : public oatpp::base::Countable {
// [...]
  template<class T>
  class Handler : public RequestHandler {
     async::Action handleError(async::Error* error) override {
        auto response = m_handler->m_controller->m_errorHandler->handleError(protocol::http::Status::CODE_500, error->what());
        return this->_return(response);
      }
    };
// [...]
public:
  std::shared_ptr<OutgoingResponse> handle(const std::shared_ptr<IncomingRequest>& request) override {
      if(m_method == nullptr) {
        if(m_methodAsync == nullptr) {
          return m_controller->handleError(Status::CODE_500, "[ApiController]: Error. Handler method is nullptr.");
        }
        return m_controller->handleError(Status::CODE_500, "[ApiController]: Error. Non-async call to async endpoint.");
      }
      if(m_controller->m_errorHandler) {
        try {
          return (m_controller->*m_method)(request);
        } catch (oatpp::web::protocol::http::HttpError& error) {
          return m_controller->m_errorHandler->handleError(error.getInfo().status, error.getMessage(), error.getHeaders());
        } catch (std::exception& error) {
          return m_controller->m_errorHandler->handleError(protocol::http::Status::CODE_500, error.what());
        } catch (...) {
          return m_controller->m_errorHandler->handleError(protocol::http::Status::CODE_500, "Unknown error");
        }
      }
      return (m_controller->*m_method)(request);
    }
// [...]
}
```
What would be the correct way to identify which error actually occurred then? Do a string match on the message? Use something other than the ErrorHandler?

My actual scenario is about the DTO validation failing with messages such as `[oatpp::parser::json::mapping::Deserializer::deserializeEnum()]: Error. Can't deserialize Enum.` in a request with hundreds of fields. I would like to identify which one failed. We assumed it would be possible to get the field name from the `oatpp::parser::ParsingError`, which does not seem possible today but could be by changing:
```cpp
namespace oatpp { namespace data { namespace mapping {
/**
 * Abstract ObjectMapper class.
 */
class ObjectMapper {
/// [...]
 template<class Wrapper>
  Wrapper readFromString(const oatpp::String& str) const {
    auto type = Wrapper::Class::getType();
    oatpp::parser::Caret caret(str);
    auto result = read(caret, type).template staticCast<Wrapper>();
    if(!result) {
      throw oatpp::parser::ParsingError( // <-- If we give it the label somehow
                                        caret.getErrorMessage(), 
                                        caret.getErrorCode(), 
                                        caret.getPosition());
    }
    return result;
  }
// [...]
}
```

### LINKED ISSUE #521: ResponseInterceptor HTTP and std exceptions are not passed as such
If a RequestInterceptor has an HTTP or std exception thrown, it is directed to the ErrorHandler. However, such an exception thrown in ResponseInterceptors don't do this. I think it should have the same behavior to have HTTP and std exceptions to also be passed in the ErrorHandler.

It looks like here is where the request interceptors HTTP and std exception is passed along to the error handler:
https://github.com/oatpp/oatpp/blob/05405fff862cb838473f4cccfb03103b54030389/src/oatpp/web/server/HttpProcessor.cpp#L117-L126

And here is where it is not being passed along for response interceptors:
https://github.com/oatpp/oatpp/blob/05405fff862cb838473f4cccfb03103b54030389/src/oatpp/web/server/HttpProcessor.cpp#L171-L177

This change could most likely be lumped in with the change of providing the entire Error to the ErrorHandler, mentioned in this comment https://github.com/oatpp/oatpp/issues/277#issuecomment-660387536.

### LINKED ISSUE #552: Hook point to handle custom exceptions
Can we add a hook point in `ApiController` to customize the behavior of the controller when there is an exception? Specifically, currently when a generic `std::exception` is thrown, the default implementation seems to just return a 500 response with `ex.what()` as the message. This is undesirable in production environment as we don't want the exception details to be visible to the end user.

I also noticed that `ApiController::handleError` is not recommended to be used yet. How about adding a customization point that looks something like this:
```cpp
class ApiController {
protected:
    // Called by endpoint when an exception is thrown during request handling
    virtual std::shared_ptr<OutgoingResponse> handleError(std::exception_ptr eptr)
    {
        // Default implementation that can be overwritten by child classes
        try {
            std::rethrow_exception(etpr);
        } catch (const oatpp::web::protocol::http::HttpError& error) {
            return m_errorHandler->handleError(error.getInfo().status, error.getMessage(), error.getHeaders());
        } catch (...) {
            return m_errorHandler->handleError(protocol::http::Status::CODE_500, "Unknown error");
        }
    }
};
```
And in a concrete controller:
```cpp
class MyController : public ApiController {
protected:
    std::shared_ptr<OutgoingResponse> handleError(std::exception_ptr eptr) override
    {
        try {
            std::rethrow_exception(etpr);
        } catch (const MyDomainException& ex) {
            return m_errorHandler->handleError(Status::CODE_500, "My domain error message");
        } catch (...) {
            // For errors we don't care, fall back to the default implementation
            return ApiController::handleError(std::current_exception());
        }
    }
};
```

JeremyGuinn:
Adds support to ErrorHandler to handle the original thrown exception.
Adds support to ApiController to assign a custom error handler at the controller level.

If the ApiController does not define an error handler, the ConnectionHandler registered error handler will handle the exception.

The changes are backwards compatible. Any currently implemented ErrorHandlers will still work since the new virtual method has a default implementation that will pass the exception info to the now deprecated methods.

The following ErrorHandler methods have been deprecated and should be removed in the next breaking release.
```c++
virtual 
std::shared_ptr<protocol::http::outgoing::Response> 
handleError(const protocol::http::Status& status, const oatpp::String& message, const Headers& headers) = 0;

std::shared_ptr<protocol::http::outgoing::Response> 
handleError(const protocol::http::Status& status, const oatpp::String& message);
```
Closes #521 
Closes #277
Closes #552

### DIFF
--- a/src/oatpp/codegen/api_controller/base_define.hpp
+++ b/src/oatpp/codegen/api_controller/base_define.hpp
@@ -70,28 +70,35 @@ const auto& OATPP_MACRO_FIRSTARG PARAM_LIST = __request;
 
 #define OATPP_MACRO_API_CONTROLLER_HEADER_1(TYPE, NAME) \
 const auto& __param_str_val_##NAME = __request->getHeader(#NAME); \
-if(!__param_str_val_##NAME){ \
-  return ApiController::handleError(Status::CODE_400, "Missing HEADER parameter '" #NAME "'"); \
+if(!__param_str_val_##NAME){                            \
+  auto httpError = oatpp::web::protocol::http::HttpError(Status::CODE_400, "Missing HEADER parameter '" #NAME "'"); \
+  auto eptr = std::make_exception_ptr(httpError); \
+  return ApiController::handleError(eptr); \
 } \
 bool __param_validation_check_##NAME; \
 const auto& NAME = ApiController::TypeInterpretation<TYPE>::fromString(#TYPE, __param_str_val_##NAME, __param_validation_check_##NAME); \
 if(!__param_validation_check_##NAME){ \
-  return ApiController::handleError(Status::CODE_400, "Invalid HEADER parameter '" #NAME "'. Expected type is '" #TYPE "'"); \
+  auto httpError = oatpp::web::protocol::http::HttpError(Status::CODE_400, "Invalid HEADER parameter '" #NAME "'. Expected type is '" #TYPE "'"); \
+  auto eptr = std::make_exception_ptr(httpError); \
+  return ApiController::handleError(eptr); \
 }
 
 #define OATPP_MACRO_API_CONTROLLER_HEADER_2(TYPE, NAME, QUALIFIER) \
 const auto& __param_str_val_##NAME = __request->getHeader(QUALIFIER); \
 if(!__param_str_val_##NAME){ \
-  return ApiController::handleError(Status::CODE_400, \
-  oatpp::String("Missing HEADER parameter '") + QUALIFIER + "'"); \
+  auto httpError = oatpp::web::protocol::http::HttpError(Status::CODE_400, oatpp::String("Missing HEADER parameter '") + QUALIFIER + "'"); \
+  auto eptr = std::make_exception_ptr(httpError); \
+  return ApiController::handleError(eptr); \
 } \
 bool __param_validation_check_##NAME; \
 const auto& NAME = ApiController::TypeInterpretation<TYPE>::fromString(#TYPE, __param_str_val_##NAME, __param_validation_check_##NAME); \
 if(!__param_validation_check_##NAME){ \
-  return ApiController::handleError(Status::CODE_400, \
+  auto httpError = oatpp::web::protocol::http::HttpError(Status::CODE_400, \
                                     oatpp::String("Invalid HEADER parameter '") + \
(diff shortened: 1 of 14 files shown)


lganzzzo (approved):
Looks great!

lganzzzo:
Merged!
Thanks for the great PR!

PR #634, opened 2022-09-28
Fix #633: Check if this extends wrapper

### LINKED ISSUE #633: Type::cast extend check is backwards
In Type::cast it checks if the Wrapper extends this but what it should be checking is if this extends the wrapper, so that casts to a parent are successful.

Could be related to https://github.com/oatpp/oatpp/issues/595

n3world:
Invert the extend check in Type::cast so that it is checking if this instance extends wrapper and therefor this can be cast to Wrapper.

### DIFF
--- a/src/oatpp/core/data/mapping/type/Type.hpp
+++ b/src/oatpp/core/data/mapping/type/Type.hpp
@@ -556,7 +556,7 @@ public:
 template <class T, class Clazz>
 template<class Wrapper>
 Wrapper ObjectWrapper<T, Clazz>::cast() const {
-  if(!Wrapper::Class::getType()->extends(m_valueType)) {
+  if(!m_valueType->extends(Wrapper::Class::getType())) {
     if(Wrapper::Class::getType() != __class::Void::getType() && m_valueType != __class::Void::getType()) {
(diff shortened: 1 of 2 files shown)


src/oatpp/core/data/mapping/type/Type.hpp
```diff
@@ -556,7 +556,7 @@ class Type {
 template <class T, class Clazz>
 template<class Wrapper>
 Wrapper ObjectWrapper<T, Clazz>::cast() const {
-  if(!Wrapper::Class::getType()->extends(m_valueType)) {
+  if(!m_valueType->extends(Wrapper::Class::getType())) {
```
lganzzzo, line 559: `ObjectWrapper<T, Clazz>::cast()` - is an unsafe static cast which allows downcasts to child classes. Thats why we check if `Wrapper::Class::getType()->extends(m_valueType)`.

n3world: I did try to use assignment copy and played around with that a little but could only seem to get raw static casts to work, which is what is being done in `StandardCollection::PolymorphicDispatcher` so maybe that is the only answer for now

Issue #608, opened 2022-08-14
Keep decimal places at 2 decimals for serializing DTO float value


DingFeng9313:
In file: oatpp/parser/json/mapping/ObjectMapper.hpp

# Question

I want to Keep decimal places at 2 decimals for serializing DTO float value, is there an API setting to do that? Or how should I work around with this requirement?

doufu3344:
I didn't find any relevant api so far.

oatpp uses the [`OATPP_FLOAT_STRING_FORMAT`](https://github.com/oatpp/oatpp/blob/bbe455af60b8f292ed90f84bd02cac1b98c2e232/src/oatpp/core/base/Config.hpp#L66)(%.16g) to format floating point data when [serialize a DTO](https://github.com/oatpp/oatpp/blob/bbe455af60b8f292ed90f84bd02cac1b98c2e232/src/oatpp/core/data/stream/Stream.cpp#L411) object. My idea is if it is sure that all conversions sharing an uniform decimal precision in you app, you can replace this macro with the format you need when compiling oatpp.

But with different precision I didn't find a proper solution using DTO.

lganzzzo:
Hello guys,

You are right - currently there is no API allowing to do this.
I think we can add something like the following:

```cpp
class MyDto : public DTO {

  DTO_INIT(MyDto, DTO)

  DTO_FIELD_INFO(value) {
    info->format = "%.16g";
  }
  DTO_FIELD(Float64, value);

};
```

emcep:
Hello. Nice to see that this issue is covered. Is there any chance this is going to be added to official release soon?

oscarthorn:
Hi! Any idea when this will be released to master?

PR #642, opened 2022-10-11
Supported for specifying float value format on serializing DTO to JSON

doufu3344:
related #608

### DIFF
--- a/src/oatpp/core/data/mapping/type/Object.hpp
+++ b/src/oatpp/core/data/mapping/type/Object.hpp
@@ -103,6 +103,10 @@ public:
       std::string pattern = "";
 
       /**
+       * Format.
+       */
+      std::string format = OATPP_FLOAT_STRING_FORMAT;
+      /**
        * Required.
        */
       bool required = false;
--- a/src/oatpp/core/data/mapping/type/Object.hpp
+++ b/src/oatpp/core/data/mapping/type/Object.hpp
@@ -105,7 +105,7 @@ public:
       /**
        * Format.
        */
-      std::string format = OATPP_FLOAT_STRING_FORMAT;
+      std::string format = "";
       /**
        * Required.
        */
--- a/src/oatpp/core/data/stream/Stream.cpp
+++ b/src/oatpp/core/data/stream/Stream.cpp
@@ -397,18 +397,18 @@ v_io_size ConsistentOutputStream::writeAsString(v_uint64 value){
   return 0;
 }
 
-v_io_size ConsistentOutputStream::writeAsString(v_float32 value){
+v_io_size ConsistentOutputStream::writeAsString(v_float32 value, const char* format){
   v_char8 a[100];
-  auto size = utils::conversion::float32ToCharSequence(value, &a[0], 100, floatFormat->c_str());
+  auto size = utils::conversion::float32ToCharSequence(value, &a[0], 100, format);
   if(size > 0){
     return writeSimple(&a[0], size);
   }
   return 0;
 }
 
-v_io_size ConsistentOutputStream::writeAsString(v_float64 value){
+v_io_size ConsistentOutputStream::writeAsString(v_float64 value, const char* format){
   v_char8 a[100];
-  auto size = utils::conversion::float64ToCharSequence(value, &a[0], 100, floatFormat->c_str());
+  auto size = utils::conversion::float64ToCharSequence(value, &a[0], 100, format);
   if(size > 0){
     return writeSimple(&a[0], size);
   }
@@ -530,6 +530,18 @@ ConsistentOutputStream& operator << (ConsistentOutputStream& s, const char* str)
   } else {
     s.writeSimple("[<char*(null)>]");
   }
+  return s;
+}
+
+template<>
+ConsistentOutputStream& operator << (ConsistentOutputStream& s, v_float32 value) {
+  s.writeAsString(value, "");
+  return s;
+}
+
+template<>
+ConsistentOutputStream& operator << (ConsistentOutputStream& s, v_float64 value) {
+  s.writeAsString(value, "");
   return s;
 }
 
--- a/src/oatpp/core/data/stream/Stream.cpp
+++ b/src/oatpp/core/data/stream/Stream.cpp
@@ -399,7 +399,7 @@ v_io_size ConsistentOutputStream::writeAsString(v_uint64 value){
 
 v_io_size ConsistentOutputStream::writeAsString(v_float32 value){
   v_char8 a[100];
-  auto size = utils::conversion::float32ToCharSequence(value, &a[0], 100);
+  auto size = utils::conversion::float32ToCharSequence(value, &a[0], 100, floatFormat->c_str());
   if(size > 0){
     return writeSimple(&a[0], size);
   }
@@ -408,7 +408,7 @@ v_io_size ConsistentOutputStream::writeAsString(v_float32 value){
 
 v_io_size ConsistentOutputStream::writeAsString(v_float64 value){
   v_char8 a[100];
-  auto size = utils::conversion::float64ToCharSequence(value, &a[0], 100);
+  auto size = utils::conversion::float64ToCharSequence(value, &a[0], 100, floatFormat->c_str());
   if(size > 0){
     return writeSimple(&a[0], size);
   }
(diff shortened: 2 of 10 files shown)


src/oatpp/core/data/mapping/type/Object.hpp
```diff
@@ -102,6 +102,10 @@ class BaseObject : public oatpp::base::Countable {
        */
       std::string pattern = "";
 
+      /**
+       * Format.
+       */
+      std::string format = OATPP_FLOAT_STRING_FORMAT;
```
lganzzzo, line 108: Let's initialize it with an empty string

```cpp
std::string format = "";
```

Then, when serializing we'll check if `format` isn't empty and choose the proper format.

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -399,7 +399,7 @@ v_io_size ConsistentOutputStream::writeAsString(v_uint64 value){
 
 v_io_size ConsistentOutputStream::writeAsString(v_float32 value){
```
lganzzzo, line 400: Let's have an additional method overload instead

```cpp
v_io_size ConsistentOutputStream::writeAsString(v_float32 value, const char* format) {
   ...
}
```

src/oatpp/core/data/stream/Stream.cpp
```diff
@@ -408,7 +408,7 @@ v_io_size ConsistentOutputStream::writeAsString(v_float32 value){
 
 v_io_size ConsistentOutputStream::writeAsString(v_float64 value){
```
lganzzzo, line 409: Let's have an additional method overload instead

```cpp
v_io_size ConsistentOutputStream::writeAsString(v_float64 value, const char* format) {
   ...
}
```

src/oatpp/core/data/stream/Stream.hpp
```diff
@@ -405,6 +405,7 @@ class IOStream : public InputStream, public OutputStream {
  */
 class ConsistentOutputStream : public OutputStream {
 public:
+  String floatFormat = OATPP_FLOAT_STRING_FORMAT;
```
lganzzzo, line 408: We don't need this since we gonna have additional methods:

```cpp
v_io_size ConsistentOutputStream::writeAsString(v_float32 value, const char* format)
v_io_size ConsistentOutputStream::writeAsString(v_float64 value, const char* format)
```

src/oatpp/parser/json/mapping/Serializer.cpp
```diff
@@ -243,7 +243,17 @@ void Serializer::serializeObject(Serializer* serializer,
       (first) ? first = false : stream->writeSimple(",", 1);
       serializeString(stream, field->name, std::strlen(field->name), serializer->m_config->escapeFlags);
       stream->writeSimple(":", 1);
+      auto streamFloatFormat = stream->floatFormat;
+      if (config->floatStringFormat != OATPP_FLOAT_STRING_FORMAT) {
+          stream->floatFormat = config->floatStringFormat;
+      }
+      if (field->info.format != OATPP_FLOAT_STRING_FORMAT) {
+          stream->floatFormat = field->info.format;
+      }
```
lganzzzo, line 252: Since we don't store `floatFormat` in the stream, we don't need this.

**Instead** - we will use the proper stream overloaded method during floatSerialization

src/oatpp/parser/json/mapping/Serializer.cpp
```diff
@@ -243,7 +243,17 @@ void Serializer::serializeObject(Serializer* serializer,
...
       serializer->serialize(stream, value);
+      if (stream->floatFormat != streamFloatFormat) {
+          stream->floatFormat = streamFloatFormat;
+      }
```
lganzzzo, line 256: Since we don't store `floatFormat` in the stream, we don't need this.

src/oatpp/parser/json/mapping/Serializer.cpp
```diff
@@ -276,6 +286,8 @@ void Serializer::serialize(data::stream::ConsistentOutputStream* stream,
 void Serializer::serializeToStream(data::stream::ConsistentOutputStream* stream,
                                    const oatpp::Void& polymorph)
 {
+  if(m_config->floatStringFormat != stream->floatFormat)
+    stream->floatFormat = m_config->floatStringFormat;
```
lganzzzo, line 290: Since we don't store `floatFormat` in the stream, we don't need this.

lganzzzo (changes requested):
Since we don't want to store floatFormat in stream we'll have to add a separate
serialization method for floats similar to [serializePrimitive](https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/mapping/Serializer.hpp#L124)
But we'll use an overloaded method from stream with the proper format

doufu3344:
I thought about this implementation before, but met some obstacles during the implementation.
I have overloaded `ConsistentOutputStream::writeAsString()` and tried a template specialization of `serializePrimitive<Float32>` looks like:
```
template<>
static void serializePrimitive<Float32>(Serializer* serializer,
                           data::stream::ConsistentOutputStream* stream,
                           const oatpp::Void& polymorph){
  (void)serializer;

  if (polymorph) {
    stream->writeAsString(*static_cast<Float32*>(polymorph.get()),
                          floatStringFormat // ???how to get the format of polymorph???
    );
  }
  else {
    stream->writeSimple("null", 4);
  }
}
```
Now, the question is: 
how to get the `floatStringFormat` in `serializePrimitive<Float32>()`, the third param(type is Float32) **holds the real value only without extra infomation**.That is, how to pass the field's format to the seializer when [serializing an object](https://github.com/oatpp/oatpp/blob/bbe455af60b8f292ed90f84bd02cac1b98c2e232/src/oatpp/parser/json/mapping/Serializer.cpp#L242-L247), i.e. how to add the format information to *value*: (serializer->serialize() --> serializePrimitive<Float32>())
https://github.com/oatpp/oatpp/blob/bbe455af60b8f292ed90f84bd02cac1b98c2e232/src/oatpp/parser/json/mapping/Serializer.cpp#L242-L247
All I can think of is something like(calling stream->writeAsString directly when field->type is floating):
```
  if (value || config->includeNullFields || (field->info.required && config->alwaysIncludeRequired)) {
    (first) ? first = false : stream->writeSimple(",", 1);
    serializeString(stream, field->name, std::strlen(field->name), serializer->m_config->escapeFlags);
    stream->writeSimple(":", 1);

    // calling stream->writeAsString directly when field->type is floating
    if (field->type == oatpp::Float32::Class::getType()) {
        stream->writeAsString(*static_cast<Float32*>(value.get()), field->info.format.c_str());
    }
    else if (field->type == oatpp::Float64::Class::getType()) {
        stream->writeAsString(*static_cast<Float64*>(value.get()), field->info.format.c_str());
    }
    else { //other types
        serializer->serialize(stream, value);
    }
  }
```
But I feel this is not perfect, is there any other solution?

lganzzzo:
@doufu3344 ,

> That is, how to pass the field's format to the seializer

Haha, that is a good question, I haven't thought of it!

The only thing that comes to my mind is to extend [serializerMethod](https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/mapping/Serializer.hpp#L118) with an extra parameter:

```cpp
  typedef void (*SerializerMethod)(Serializer*,
                                   data::stream::ConsistentOutputStream*,
                                   const oatpp::Void&,
                                   void* extra);
```

Or even better - like this:

```cpp

  struct Context {
    Serializer* serializer;
    data::mapping::type::BaseObject::Property::Info* info;
    // we can add more later if needed
  }

  typedef void (*SerializerMethod)(const std::unique_ptr<Context>&, // <-- Now we pass Context instead
                                   data::stream::ConsistentOutputStream*,
                                   const oatpp::Void&);

```

Then [here](https://github.com/oatpp/oatpp/blob/master/src/oatpp/parser/json/mapping/Serializer.cpp#L246) we do:

```cpp
context->info = &field->info;
serializer->serialize(context, value);
```

Context is created once on serialization start and destroyed on serialization end.

lganzzzo (approved):

lganzzzo:
Merged to [#642_candidate](https://github.com/oatpp/oatpp/tree/%23642_candidate) branch.
After some more testing will be merged to master.
Great work!

PR #660, opened 2022-11-17
allow readBodyToStringAsync to support empty body

doufu3344:
A null string can properly get using macro BODY_STRING at sync mode, but at async mode, an exception is thrown and the next processing flow is interrupted. Sync and async mode should be in a unified way.


lganzzzo:
Fixed in #702 
I'm closing this PR.

Issue #662, opened 2022-11-24
    Chinese is not supported


cyang-catalogicsoftware:
![image](https://user-images.githubusercontent.com/101167617/203693929-fc51179b-58a3-4b89-a3b5-78c11c6bbed4.png)

Data received by the browser
![image](https://user-images.githubusercontent.com/101167617/203693993-b4ad3abd-a2a1-413a-9942-c7fda6027f8e.png)

Can you help me?  thanks

SSDGADsss:
JavaScript use UTF-16 charset , You can try "std::wstring_convert" this function, convert UTF-8 to UTF-16

cyang-catalogicsoftware:
I tested and found that the problem still exists

![image](https://user-images.githubusercontent.com/101167617/203945130-9658b1dc-5c7f-447e-b0d5-9ce68fb58b73.png)

Can you verify it  ?  thanks

cyang-catalogicsoftware:
I really need your help

doufu3344:
I think it's a browser decoding problem: the browser doesn't know the charset of the response causes displaying error.

1. make sure your string literal is encoded by utf-8, you source file's encoding is utf-8 or adding [`u8`](https://en.cppreference.com/w/cpp/language/string_literal) prefix to the literal.
2. add `Content-Type` header to your response with the correct charset. If you have many endpoints, you can reference [ResponseInterceptor](https://github.com/oatpp/oatpp/blob/master/changelog/1.2.5.md#introduce-responseinterceptor) to and this header globally.

![Snipaste_2022-11-28_17-13-24](https://user-images.githubusercontent.com/1912467/204244675-a6fe18fb-285c-4375-9ec4-73d6636a371e.png)

![Snipaste_2022-11-28_17-10-47](https://user-images.githubusercontent.com/1912467/204244692-c1ca0764-ea07-4d46-9c63-54d721c080dc.png)

cyang-catalogicsoftware:
Thank you very much , Are you Chinese?

doufu3344:
yep

cyang-catalogicsoftware:
![image](https://user-images.githubusercontent.com/101167617/204482648-1442f656-129b-49ac-8466-5812e3718552.png)

这种情况该如何处理呢？  如果返回的是dto，编码还是会出问题

cyang-catalogicsoftware:
[...]

PR #667, opened 2022-11-29
Support for disabling escaping of non-ASCII when serializing DTOs

doufu3344:
non-ASCII(utf8) characters can output as it's origin bytes other than an unicode encoding form.

#662


lganzzzo:
Looks good
Merged!

PR #680, opened 2022-12-09
Fix parsing of Boolean

### LINKED ISSUE #673: Error when parsing a Boolean
I think there might be a bug in `ObjectMapper<Boolean>::readFromString`:
https://github.com/oatpp/oatpp/blob/6063b57b250bfa9bdf02c47351a4a4ce1a969079/src/oatpp/core/data/mapping/ObjectMapper.hpp#L124-L133

When the given string is `"false"`, it is correctly parsed, but the `bool` conversion operator of `Boolean`
https://github.com/oatpp/oatpp/blob/6063b57b250bfa9bdf02c47351a4a4ce1a969079/src/oatpp/core/data/mapping/type/Primitive.hpp#L431-L436

gets invoked at line 129 and an error is raised. The program then crashes because `caret.getErrorMessage()` is null.

Reproduction:
```cpp
#include "oatpp/parser/json/mapping/ObjectMapper.hpp"
#include "gtest/gtest.h"

using namespace oatpp::parser::json::mapping;

TEST(oatpp, parse_true)
{
   ObjectMapper json_mapper(Serializer::Config::createShared(),
                            Deserializer::Config::createShared());
   ASSERT_TRUE(json_mapper.readFromString<oatpp::Boolean>("true"));
}

TEST(oatpp, parse_false)
{
   ObjectMapper json_mapper(Serializer::Config::createShared(),
                            Deserializer::Config::createShared());
   ASSERT_FALSE(json_mapper.readFromString<oatpp::Boolean>("false")); // <- crash
}
```

Inujel:
Fixes #673

### DIFF
--- a/src/oatpp/core/data/mapping/ObjectMapper.hpp
+++ b/src/oatpp/core/data/mapping/ObjectMapper.hpp
@@ -126,7 +126,7 @@ public:
     auto type = Wrapper::Class::getType();
     oatpp::parser::Caret caret(str);
     auto result = read(caret, type).template cast<Wrapper>();
-    if(!result) {
+    if(result == nullptr) {
       throw oatpp::parser::ParsingError(caret.getErrorMessage(), caret.getErrorCode(), caret.getPosition());
     }
     return result;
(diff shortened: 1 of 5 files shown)


lganzzzo:
Looks good!
Merged

PR #682, opened 2022-12-11
closes #681

### LINKED ISSUE #681: Prettify CMake
Hi,
modern CMake can scan files in a folder recursively and mirror the folder structure in the project hierarchy.
So if you add/rename/remove source files you don't need to touch the includes in Cmake :-)
Also, the code can be browsed much more easily. I am using Visual Studio and for me it is pretty much a relief with these changes.
All we need is CMake >= 3.8 which has been out for more than five years now.

DanielHerchenbach:
I am using CMake command
file with GLOB_RECURSE to scan source files
and then
source_group with TREE to place them in a hierarchy for better overview in the IDE. Those commands should work with CMake version >= 3.8.


lganzzzo:
Hello @DanielHerchenbach ,

Please wait for PR approvals 🤝  :)

lganzzzo:
Hello @bamkrs , what do you think about globbing instead of source listing? Any objections?

DanielHerchenbach:
Ups, I just found this on Stack Overflow:
[https://cmake.org/cmake/help/latest/command/file.html?highlight=glob#glob](https://cmake.org/cmake/help/latest/command/file.html?highlight=glob#glob) (refer to the note)
Do you use a build system that actually checks whether any CMakeLists.txt file has changed?

lganzzzo:
CI builds everything from scratch.
And I recommend to clear CMake cache when new files added so it shouldn't be much of a problem.
However, as for me, globbing isn't much of an improvement either:)

Let's wait for @bamkrs comment.

PR #719, opened 2023-03-05
Fix possible race condition on destroy

i96751414:
This is just for correctness - I did not have any problems with destroy so far.


lganzzzo (approved):
Looks good

lganzzzo:
Merged!

PR #753, opened 2023-05-10
CMake source groups

sodevel:
This reverts #682 and restores the recommended CMake style of explicit listing of all source files of a target. Source groups are still added like #682 did, but this is done by querying the targets source files. Additionally, multiple source groups are used like Visual Studio does by default.

This increases the required CMake version to 3.20.


lganzzzo (approved):
Looks good

lganzzzo:
Merged!

sodevel:
Like all previous posts like this before, this doesn't make sense and should not be possible at all. Again a regular source file is added to the generated files group which is wrong. Your used CMake version allows to run the configure phase under a debugger (this is very easy to do with VS Code by selecting the CMake tab on the sidebar and click on the bug icon in the header), can u place a break point on line 107 in the file `cmake/project.cmake` and debug the configure run?

The third time it breaks is right before your error, i am interested in the values of the CMake variables `fileSources`, `sourceTreeDir`, `binaryTreeDir` and `binaryFiles` (in my VS Code version i somehow can't copy values from the Variables group, i have to add them as Watch and copy their values from that group).

sodevel:
I don't think this is the reason, this is a common situation and shouldn't cause an issue. What i do not understand is, why can't i trigger that error for me? Why doesn't it happen in my other projects?

The other issues #804 and #844 have one thing in common with your issue, the error happens in https://github.com/oatpp/oatpp/blob/b85a565351ec65fde5b24c5b1ba2b89d515c5439/cmake/project.cmake#L106-L111

And there, something not so common does happen, due to the missing quoting the variable `filterSources` actually gets removed. Maybe this triggers some CMake bug and causes this error? Can you change line 106 to
```cmake
set(filterSources "${binaryFiles}")
```
and check if this fixes the error for you?

sodevel:
Thank you, thats the reason why i can't trigger the error, i was using CMake 3.27.7. I have created an [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/25353) for CMake, they also think this is a regression.

PR #763, opened 2023-05-23
Fix CMake regular expression injection

### LINKED ISSUE #761: CMake regex bug
```
10:06:11 oatpp version: '1.3.0'
10:06:11 -- MSVC: using dynamically-linked runtime (/MD and /MDd).
10:06:11 CMake Error at cmake/project.cmake:66 (list):
10:06:11   list sub-command FILTER, mode REGEX failed to compile regex
10:06:11   "^D:/workspace/Oat++/build/src/".
10:06:11 Call Stack (most recent call first):
10:06:11   src/CMakeLists.txt:297 (set_target_source_groups)
```

CMake: 3.26.4
Source: 7e520ec646574962521f076ef16f22527e9f289d

If you have special characters like + or - in the file path, the CMake script will not work!

Fix:
```
diff --git a/cmake/project.cmake b/cmake/project.cmake
index 7bd25c49..c97deb67 100644
--- a/cmake/project.cmake
+++ b/cmake/project.cmake
@@ -1,5 +1,10 @@
 cmake_minimum_required(VERSION 3.20)

+macro(fixregexpath pathvar)
+  string(REPLACE "+" "\\+" ${pathvar} ${${pathvar}})
+  string(REPLACE "-" "\\-" ${pathvar} ${${pathvar}})
+endmacro()
+
 #[[
 Set default source groups.

@@ -62,12 +67,14 @@ function(set_target_source_groups arg_TARGET)
   # Normally, the build tree is a subdirectory of the source tree. For the root directory the prefix match
   # will also match the binary tree, so create separate file lists by filtering the binary prefix explicit.
   # This will fail when doing an in-tree build, in that case everything will be classified as binary.
+  fixregexpath(binaryDir)
   set(sourceFiles ${fileSources})
   list(FILTER sourceFiles EXCLUDE REGEX "^${binaryDir}/")
   set(binaryFiles ${fileSources})
   list(FILTER binaryFiles INCLUDE REGEX "^${binaryDir}/")

   set(filterSources ${sourceFiles})
+  fixregexpath(sourceDir)
   list(FILTER filterSources INCLUDE REGEX "^${sourceDir}/.+\\.h(h|pp)?$")
   source_group(
     TREE "${sourceTreeDir}"
```

sodevel:
Don't insert paths into regular expressions,
paths can contain regular expression meta characters which will get evaluated.
Process file lists manually and use cmake_path(IS_PREFIX) to filter paths.

Closes #761

### DIFF
--- a/cmake/project.cmake
+++ b/cmake/project.cmake
@@ -75,28 +75,28 @@ function(set_target_source_groups arg_TARGET)
   endforeach()
 
   set(filterSources ${sourceFiles})
-  list(FILTER filterSources INCLUDE REGEX "^${sourceDir}/.+\\.h(h|pp)?$")
+  list(FILTER filterSources INCLUDE REGEX "/.+\\.h(h|pp)?$")
   source_group(
     TREE "${sourceTreeDir}"
     PREFIX "Header Files"
     FILES ${filterSources}
   )
   set(filterSources ${sourceFiles})
-  list(FILTER filterSources INCLUDE REGEX "^${sourceDir}/.+\\.h(h|pp)?\\.in$")
+  list(FILTER filterSources INCLUDE REGEX "/.+\\.h(h|pp)?\\.in$")
   source_group(
     TREE "${sourceTreeDir}"
     PREFIX "Header Templates"
     FILES ${filterSources}
   )
   set(filterSources ${sourceFiles})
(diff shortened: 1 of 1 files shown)


lganzzzo (approved):

lganzzzo:
Merged!

Issue #784, opened 2023-07-21
Lots of compiler warnings


fhuberts:
Hi,

We've integrated oatpp into our application nicely.
However, when we leave all our compiler flags on we get loads of compiler warnings.
To get thing to compile at all, and without warnings we have to disable a lot of warning.
The below is taken from our CMake files.
```
  # disable the warnings that oat++ can't handle
  remove_compiler_flags("-Wextra-semi")
  remove_compiler_flags("-Wold-style-cast")
  remove_compiler_flags("-Wshadow")
  remove_compiler_flags("-Wctor-dtor-privacy")
  remove_compiler_flags("-Wuseless-cast")
  remove_compiler_flags("-Wdouble-promotion")
  remove_compiler_flags("-Wsign-conversion")
  remove_compiler_flags("-Wswitch-default")
  remove_compiler_flags("-Wmissing-declarations")
  remove_compiler_flags("-Wredundant-tags")
  remove_compiler_flags("-Warith-conversion")
  remove_compiler_flags("-Wswitch-enum")
  remove_compiler_flags("-Wunused-parameter")
  remove_compiler_flags("-Wsuggest-override")
  remove_compiler_flags("-Wduplicated-branches")
  remove_compiler_flags("-Wlogical-op")
  remove_compiler_flags("-Wsuggest-attribute=cold")
  remove_compiler_flags("-Wsuggest-attribute=noreturn")
  remove_compiler_flags("-Wconversion")
  add_compiler_flags("-Wno-suggest-attribute=format")
  add_compiler_flags("-Wno-sign-compare")
  add_compiler_flags("-Wno-pessimizing-move")
  add_compiler_flags("-Wno-unused-parameter")
```

Could you please fix these?

Some are quite harmless, like the 'extra-semi', but others most probably hide bugs, like 'sign-conversion'.

We're compiling oatpp on Debian 12 x64 and Fedora 38 x64.

fhuberts:
Submitted 4 PRs. Will submit more after these.

fhuberts:
After the 7 PRs that are now submitted, we still have a few flags left:
- -Wsign-conversion: fixing this is a huge work, many many warnings and lots of casting and careful review required. I have a commit that at least fixes these warnings in headers so that projects using oatpp do not need to ignore these warnings.
- -Wunused-parameter: fixing this one requires updating to c++17. Any chance? 2017 was 6 years ago ;-)
[...]

PR #788, opened 2023-07-22
Fix compiler warnings

fhuberts:
Some more fixes.


fhuberts:
will submit PR per commit.

I hate windows...

PR #789, opened 2023-07-22
Fix compiler warnings (-Wsuggest-override)

fhuberts:


fhuberts:
I have no windows machine, so if anyone can help me fix this build that would be much appreciated.

lganzzzo (approved):

lganzzzo:
Hello @fhuberts ,

> I have no windows machine, so if anyone can help me fix this build that would be much appreciated.

I don't see anything in this PR that could break the build.
If it's not some random glitch - then good that build is failing.

lganzzzo:
Merged

PR #792, opened 2023-07-22
Fix compiler warnings (-Wunused-parameter)

fhuberts:


fhuberts:
I have no windows machine, so if anyone can help me fix this build that would be much appreciated.

lganzzzo:
Hello @fhuberts ,

`[[maybe_unused]]` is C++ 17 feature. We have not yet raised the min-supported C++ version to 17.
Current min supported cpp is 11.

fhuberts:
ok, tnx.
I see that I completely missed that in the cmake file, sorry.

PR #806, opened 2023-08-07

fhuberts:
See #784

### DIFF
--- a/src/oatpp/network/tcp/server/ConnectionProvider.cpp
+++ b/src/oatpp/network/tcp/server/ConnectionProvider.cpp
@@ -415,7 +415,7 @@ provider::ResourceHandle<oatpp::data::stream::IOStream> ConnectionProvider::get(
     timeout.tv_sec = 1;
     timeout.tv_usec = 0;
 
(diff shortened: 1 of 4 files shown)


src/oatpp/network/tcp/server/ConnectionProvider.cpp
```diff
@@ -415,7 +415,7 @@ provider::ResourceHandle<oatpp::data::stream::IOStream> ConnectionProvider::get(
     timeout.tv_sec = 1;
     timeout.tv_usec = 0;
 
-    auto res = select(int(m_serverHandle + 1), &set, nullptr, nullptr, &timeout);
+    auto res = select(m_serverHandle + 1, &set, nullptr, nullptr, &timeout);
```
lganzzzo, line 418: On windows `v_io_handle` is an unsigned int,- this will trigger a different warning.

lganzzzo (commented):
In general I would suppress `-Wuseless-cast` instead of following its fixes.
It doesn't seem useful in anyway to me.

fhuberts:
Just discovered that this one is dependent on the old-style-cast commit.
will fix the windows warning are resubmit later.

PR #810, opened 2023-08-12
Fix multiple warnings

fhuberts:
See #784 

I've made a PR with multiple warning fixes to make it a bit more clear what I still have open for submission.

If you want me want me to submit them individually, then please tell me.
Or you could just cherry pick the commits after you've reviewed them

Do note that there is some overlap between the commit, so I'd advise that the order is preserved (just to make it bit easier to apply).


fhuberts:
I've rebased this branch with __attribute__ fixes for Windows builds

fhuberts:
Made the changes to Caret in 'Fix compiler warnings (-Wconversion)' safer (forgot I had these pending as well)

src/oatpp/compiler.hpp
```diff
@@ -0,0 +1,34 @@
...
+ *
+ ***************************************************************************/
+
+#ifndef oatpp_compiler_hpp
+#define oatpp_compiler_hpp
+
+#if defined(WIN32) || defined(_WIN32)
+  #define COMPILER_ATTRIBUTE(x)
+#else
+  #define COMPILER_ATTRIBUTE(x) __attribute__((x))
```
lganzzzo, line 31: Non-standard sugar (GCC specific attributes) isn't allowed in "Vanilla" oatpp code.
We can't add this file.

src/oatpp/core/async/worker/IOEventWorker_epoll.cpp
```diff
@@ -162,8 +162,8 @@ void IOEventWorker::waitEvents() {
     OATPP_LOGE("[oatpp::async::worker::IOEventWorker::waitEvents()]", "Error:\n"
                "errno=%d\n"
                "in-events=%d\n"
-               "foreman=%d\n"
-               "this=%d\n"
+               "foreman=%p\n"
```
lganzzzo, line 165: Please use `%llu` or `%llx` for pointers with cast to `v_ubuff_size`.
To have unified output on all platforms

src/oatpp/network/tcp/Connection.cpp
```diff
@@ -115,7 +115,7 @@ v_io_size Connection::write(const void *buff, v_buff_size count, async::Action&
 
   if(result < 0) {
     auto e = errno;
-    if(e == EAGAIN || e == EWOULDBLOCK){
+    if(e == EAGAIN){ // EAGAIN === EWOULDBLOCK
```
lganzzzo, line 118: Please revert this back to 
`if(e == EAGAIN || e == EWOULDBLOCK) ...`

We assume that somewhere may exist exotic platform where `EAGAIN` and `EWOULDBLOCK` may have different values

lganzzzo (changes requested):

src/oatpp/compiler.hpp
fhuberts: That is very unfortunate.
This means that I can't add the attributes and that the compiler can't catch errors in usage of the functions, like it did in UrlTest.cpp.

How about if I make the code a bit more specific?

```
#if __GNUC__
  #define COMPILER_ATTRIBUTE(x) __attribute__((x))
#else
  #define COMPILER_ATTRIBUTE(x)
#endif
```

I would really like to make a case for adding these attributes, because they catch actual bugs.

src/oatpp/network/tcp/Connection.cpp
fhuberts: I know of no platform that has these as different values.
I'll try to come up with a solution.
If that is not possible then fixing -Wlogical-op is not possible.

src/oatpp/core/async/worker/IOEventWorker_epoll.cpp
fhuberts: done

src/oatpp/network/tcp/Connection.cpp
fhuberts: I think the solution I came up is acceptable?

fhuberts:
I will rebase this after #816 is merged if that is ok.

fhuberts:
Went ahead and rebased it on #816, and folded the follow-up patches into the proper commits

src/oatpp/compiler.hpp
lganzzzo: It might work.
Let's place this file in `src/oatpp/core/base/Compiler.hpp`.
I'll think of a better place for it later

src/oatpp/network/tcp/Connection.cpp
lganzzzo: Looks like we can do the following:

```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wlogical-op "

/* code */

#pragma GCC diagnostic pop
```

src/oatpp/compiler.hpp
fhuberts: ok

fhuberts: this is now in its own PR

src/oatpp/network/tcp/Connection.cpp
fhuberts: will do that in its own PR.

fhuberts:
I have now submitted the first 4 commits of this series as separate PRs.
That makes discussing about the commits a lot easier and will probably resolve the discussions quicker.
Those 4 PRs are commits that are not order-dependent.

After resolving and merging those I'll continue with the rest of the series, and incorporate your comments into them.

I think this PR can be closed now.

lganzzzo:
All 5 new PRs were merged.
Should we close this one?

fhuberts:
thanks for merging those 5, more are coming, but I'll close this one.

PR #811, opened 2023-08-12

fhuberts:
See #784 

The first in the pending series.


fhuberts:
closing, is part of #810

PR #812, opened 2023-08-12
Fix wsuggest attribute format

fhuberts:
See #784

On top of the previous PR


fhuberts:
closing, is part of #810

PR #816, opened 2023-08-14
Compiler flags are for gcc

fhuberts:
See #784 

This is a small part of a series I'm working on to add compiler flags without testing them (when it's a known supported flag). Not testing known good flags makes the build setup much faster (after the slowdown that was introduced by me adding the flags...).

I also have a list of Clang flags that can be added if needed/wanted, let me know.


fhuberts:
The series is now complete.
It speeds up the build quite nicely without losing functionality.

If you want, I can add Clang warning flags in a separate PR.

lganzzzo (approved):

lganzzzo:
merged

PR #826, opened 2023-08-17
Fix compiler warnings (-Wlogical-op)

fhuberts:
See #784

### DIFF
--- a/src/oatpp/network/tcp/Connection.cpp
+++ b/src/oatpp/network/tcp/Connection.cpp
@@ -115,7 +115,19 @@ v_io_size Connection::write(const void *buff, v_buff_size count, async::Action& 
 
   if(result < 0) {
     auto e = errno;
-    if(e == EAGAIN || e == EWOULDBLOCK){
+
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wlogical-op"
+#endif
+
+    bool retry = ((e == EAGAIN) || (e == EWOULDBLOCK));
+
+#ifdef __GNUC__
+#pragma GCC diagnostic pop
+#endif
+
+    if(retry){
       if(m_mode == data::stream::ASYNCHRONOUS) {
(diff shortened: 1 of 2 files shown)


src/oatpp/network/tcp/Connection.cpp
```diff
@@ -115,7 +115,19 @@ v_io_size Connection::write(const void *buff, v_buff_size count, async::Action&
 
   if(result < 0) {
     auto e = errno;
-    if(e == EAGAIN || e == EWOULDBLOCK){
+
```
lganzzzo, line 118: Let's suppress warning for the whole function - 
just to make code more readable.

```cpp
#ifdef __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wlogical-op"
#endif

function {
 ...
}

#ifdef __GNUC__
#pragma GCC diagnostic pop
#endif

```

src/oatpp/network/tcp/Connection.cpp
```diff
@@ -173,7 +185,19 @@ v_io_size Connection::read(void *buff, v_buff_size count, async::Action& action)
...
```
lganzzzo, line 188: Let's suppress warning for the whole function - 
just to make code more readable.

```cpp
#ifdef __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wlogical-op"
#endif

function {
 ...
}

#ifdef __GNUC__
#pragma GCC diagnostic pop
#endif

```

lganzzzo (changes requested):

src/oatpp/network/tcp/Connection.cpp
fhuberts: done

src/oatpp/network/tcp/Connection.cpp
fhuberts: done

lganzzzo (approved):

lganzzzo:
merged

PR #829, opened 2023-08-17

fhuberts:
The compiler didn't catch this one in the oatpp build, but did catch it in the project in which we integrated oatpp.

See #784

### DIFF
--- a/src/oatpp/core/data/mapping/type/Type.cpp
+++ b/src/oatpp/core/data/mapping/type/Type.cpp
@@ -68,7 +68,7 @@ int ClassId::getClassCount() {
 
 std::vector<const char*> ClassId::getRegisteredClassNames() {
   std::lock_guard<std::mutex> lock(getClassMutex());
-  return std::vector<const char*>(getClassNames());
+  return getClassNames();
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


fhuberts:
I'm still getting this warning on Debian 12 x64, but I'm unsure how to fix it.
Can you please take a look?

```
10:17:39  oatpp/src/oatpp/core/data/mapping/type/Type.cpp: In static member function ‘static std::vector<const char*> oatpp::data::mapping::type::ClassId::getRegisteredClassNames()’:
10:17:39  oatpp/src/oatpp/core/data/mapping/type/Type.cpp:71:15: warning: useless cast to type ‘class std::vector<const char*>’ [-Wuseless-cast]
10:17:39     71 |   return std::vector<const char*>(getClassNames());
10:17:39        |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

lganzzzo:
> I'm still getting this warning on Debian 12 x64, but I'm unsure how to fix it (on Fedora 38 x64 I have no warnings).
Can you please take a look?

Debian is right

```cpp
std::vector<const char*>& ClassId::getClassNames() {
  static std::vector<const char*> classNames;
  return classNames;
}

...

std::vector<const char*> ClassId::getRegisteredClassNames() {
  std::lock_guard<std::mutex> lock(getClassMutex());
  return std::vector<const char*>(getClassNames());
}
```

`getRegisteredClassNames()` can be simplified:

```cpp
std::vector<const char*> ClassId::getRegisteredClassNames() {
  std::lock_guard<std::mutex> lock(getClassMutex());
  return getClassNames();
}
```

fhuberts:
Added the fix

fhuberts:
I don't know why the Windows build fails.

lganzzzo (approved):

lganzzzo:
merged

PR #832, opened 2023-08-23
Fix issue with "100 Continue" response not sent correctly

### LINKED ISSUE #607: oatpp handles "Expect: 100-continue" incorrectly
When curl sends "Expect" 100-continue", oatpp sends "100 Continue" response back. However, it's not a full response. It's just the status line "HTTP/1.1 100 Continue CRLF". But for the full (body-less) response you need another CRLF.

I tested with curl. Some curl versions (curl 7.48) hang after the main response, some (curl 7.56) proceed. But all ignore the "100 Continue" and wait for 1s to continue with body transfer. I know that the 1s wait is part of curl logic, but it does indicate it ignores the response. The older version I mentioned (curl 7.48) seems to lose 2 bytes of the main response and waits for more. I suspect those 2 bytes are related to the missing CRLF from oatpp. Maybe it just assumes the additional CRLF will always be there. And the newer versions of curl check that more rigorously.

After I added additional "\r\n" to the RESPONSE_100_CONTINUE: "HTTP/1.1 100 Continue\r\n\r\n" in SimpleBodyDecoder.cpp, all curl versions stopped hanging on the main response, finishing correctly. Also, they stopped waiting for 1s, properly processing the "100 Continue" response and commencing the body transfer right away.

The RFC seems to indicate that even informational responses must be full responses, hence I think the additional CRLF is in conformance.

patrikhuss:
fixes #607

### DIFF
--- a/src/oatpp/web/protocol/http/incoming/SimpleBodyDecoder.cpp
+++ b/src/oatpp/web/protocol/http/incoming/SimpleBodyDecoder.cpp
@@ -31,7 +31,7 @@
 
 namespace oatpp { namespace web { namespace protocol { namespace http { namespace incoming {
 
-const std::string SimpleBodyDecoder::RESPONSE_100_CONTINUE = "HTTP/1.1 100 Continue\r\n";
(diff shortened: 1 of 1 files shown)


lganzzzo (approved):

lganzzzo:
Merged!

PR #835, opened 2023-08-24
Fix compiler warnings in headers (-Wsign-conversion)

fhuberts:
See #784

### DIFF
--- a/src/oatpp/core/data/mapping/type/Collection.hpp
+++ b/src/oatpp/core/data/mapping/type/Collection.hpp
@@ -166,7 +166,7 @@ public:
 
     v_int64 getCollectionSize(const type::Void& object) const override {
       ContainerType* collection = static_cast<ContainerType*>(object.get());
-      return collection->size();
+      return static_cast<v_int64>(collection->size());
     }
 
     void addItem(const type::Void& object, const type::Void& item) const override {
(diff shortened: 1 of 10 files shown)


fhuberts:
rebased on #837

src/oatpp/web/server/handler/ErrorHandler.cpp
```diff
@@ -32,8 +32,10 @@ std::shared_ptr<protocol::http::outgoing::Response> ErrorHandler::handleError(co
 
   std::shared_ptr<protocol::http::outgoing::Response> response;
 
+#ifdef __GNUC__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
lganzzzo, line 38: @fhuberts , I'm not sure if we should hide deprecated methods usage - instead we want it to be fixed.

fhuberts: Yeah I agree, but in one build it fails on the pragmas, and then in another build it also fails when I hid the pragmas.

So I don't know what to do.

fhuberts: I've pulled out the commit "Wrap GCC pragma in '#ifdef __GNUC__'"

lganzzzo (approved):

lganzzzo:
merged

fhuberts:
thanks!

PR #837, opened 2023-08-24
Wrap GCC pragma in '#ifdef __GNUC__'

fhuberts:
See #836


fhuberts:
Closing because of your remarks in #835

PR #842, opened 2023-08-31
Fix compiler warnings in compilation units (-Wsign-conversion)

fhuberts:
re-opened from #836


src/oatpp/parser/json/mapping/Deserializer.cpp
```diff
@@ -494,7 +494,7 @@ oatpp::Void Deserializer::deserializeObject(Deserializer* deserializer, parser::
 }
 
 oatpp::Void Deserializer::deserialize(parser::Caret& caret, const Type* const type) {
-  auto id = type->classId.id;
+  v_uint32 id = static_cast<v_uint32>(type->classId.id);
```
lganzzzo, line 497: We can use `auto` here

src/oatpp/parser/json/mapping/Serializer.cpp
```diff
@@ -259,7 +259,7 @@ void Serializer::serializeObject(Serializer* serializer,
 void Serializer::serialize(data::stream::ConsistentOutputStream* stream,
                            const oatpp::Void& polymorph)
 {
-  auto id = polymorph.getValueType()->classId.id;
+  v_uint32 id = static_cast<v_uint32>(polymorph.getValueType()->classId.id);
```
lganzzzo, line 262: auto here:

```cpp
auto id = static_cast<v_uint32>(polymorph.getValueType()->classId.id);
```

lganzzzo (commented):
Hello @fhuberts ,

Looks good to me.
Please fix two minor typos

Also, I have a minor concern about portability of `socklen_t `.
It might be a good idea if we add oatpp specific type for that (/core/IODefinitions.hpp):

Something like:

```cpp
#ifdef _WIN32
    typedef int v_sock_size; 
#else
    typedef socklen_t v_sock_size;
#endif
```

fhuberts:
Done

src/oatpp/parser/json/mapping/Deserializer.cpp
fhuberts: done

src/oatpp/parser/json/mapping/Serializer.cpp
fhuberts: done

lganzzzo (approved):

lganzzzo:
Merged!

PR #849, opened 2023-09-14
use free lock queue to optimize DefaultLogger

linhaojun857:
- I also wrote the test code of FreelackQueue，and I have tested a lot in the local area, which is also OK
- The following is the time comparison of output 10 million logs
  - Before optimization

    before-optimization

  - After optimization

    after-optimization


linhaojun857:
@lganzzzo  Hi, Can you help me do a code review for this pull request?

lganzzzo:
Hello @linhaojun857 ,

Thanks for the PR!

> @lganzzzo Hi, Can you help me do a code review for this pull request?

I've seen it. Just need some more time to review and test it

PR #853, opened 2023-09-15
Fix client retry bug

### LINKED ISSUE #770: RetryPolicy max attempts doesn't matter when connection handle cannot be obtained
Hi,

It appears that when `web::client::RequestExecutor::executeAsync` encounters a situation where `ExecutorCoroutine::m_connectionHandle` cannot be obtained (e.g., when the `getaddrinfo` call fails), `ExecutorCoroutine::retry` is invoked more times than allowed by the RetryPolicy, if given any.

This could potentially lead to an infinite number of retries without ever reaching the user's `handleError` method.

I guess `ExecutorCoroutine::act` should increment the `m_context.attempt` variable instead of `ExecutorCoroutine::execute`.

Best regards

linhaojun857:
fix #770

### DIFF
--- a/src/oatpp/web/client/RequestExecutor.cpp
+++ b/src/oatpp/web/client/RequestExecutor.cpp
@@ -150,6 +150,7 @@ RequestExecutor::executeAsync(
     {}
 
     Action act() override {
+      m_context.attempt ++;
 
       if(!m_connectionHandle) {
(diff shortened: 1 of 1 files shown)


linhaojun857:
fix #770

linhaojun857:
This ensures every retry can  ***m_context.attempt++***

lganzzzo (approved):
Looks good!

PR #855, opened 2023-09-16
Add Interface ConnectionConfigurer

### LINKED ISSUE #775: TCP_NODELAY
Hi!

I'm looking to set some tcp socket options, TCP_NODELAY and maybe TCP_QUICKACK when using websockets. Is there a way to do this with oat++? I found the socket in `oatpp::network::tcp::server::ConnectionProvider` but since the variable is private the is no easy way to set these options without rewriting the class.

linhaojun857:
fix #775
User can implement `ConnectionConfigurer`, and add it to `ConnectionProvider` by `setConnectionConfigurer` method to achieve the setting of the Socket options


lganzzzo (approved):

lganzzzo:
Nice!
Merged!

PR #861, opened 2023-09-21
Fix serialize object and deserialize json when not include required field

### LINKED ISSUE #676: Throwing exception when trying to deserialize JSON which does not include field marked as required
DTO fields may be marked as `required` (see #417) but this information is ignored while deserializing JSON using `readFromString`.

Do you find it useful to throw an exception when JSON does not contain a key marked as `required` in corresponding DTO? Of course, throwing should be optional.
IMO it would be helpful, as now in my project I have to write separate functions for each DTO which validates if all fields I need are not nullptrs after deserialization.

I'm just wondering what do you think about this idea before trying to create a PR with changes.

Maybe something similar already exists? For given object, can I access the values of all required fields? I know that I may check which fields are required
```
for (const auto [name, property] : dtoWrapped.getPropertiesMap())
{
    // property->info.required;
}
```
Is it possible to access the value of `dtoWrapped`'s member called `name` in this loop?
If so, then I may write one generic function which checks if all fields marked as `required` are initialized (instead of writing separate functions for each DTO class).

linhaojun857:
fix #676

### DIFF
--- a/src/oatpp/parser/json/mapping/Deserializer.cpp
+++ b/src/oatpp/parser/json/mapping/Deserializer.cpp
@@ -447,7 +447,13 @@ oatpp::Void Deserializer::deserializeObject(Deserializer* deserializer, parser::
           skipValue(caret);
           polymorphs.emplace_back(field, label.toString()); // store polymorphs for later processing.
         } else {
-          field->set(static_cast<oatpp::BaseObject *>(object.get()), deserializer->deserialize(caret, field->type));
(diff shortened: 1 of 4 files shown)


linhaojun857:
@lganzzzo Hello, Can you help me do a code review for this pull request?

lganzzzo (approved):

lganzzzo:
Thanks for the PR. Looks Good!
Merged!

PR #863, opened 2023-09-22
Add option to customize error message for validating input

### LINKED ISSUE #733: customize message for validating input
Hi,

I have an interesting problem,

class MyDto : public oatpp::DTO {

  DTO_INIT(MyDto, DTO)

  DTO_FIELD(String, abc);
  DTO_FIELD(Int32, def);

 static oatpp:object....{
     auto dto = createShared();
    return dto
}
};

If user inserts wrong type input, is there anyway we can customize message instead of OAT internal message error? e.g. if user insert string type for def, we can show a message "**def should be integer format**" instead of "ERROR_INVALID_INTEGER"

linhaojun857:
fix #733

### DIFF
--- a/src/oatpp/core/data/mapping/type/Object.hpp
+++ b/src/oatpp/core/data/mapping/type/Object.hpp
@@ -103,6 +103,11 @@ public:
       std::string pattern = "";
 
       /**
+       * Error message.
+       */
+      std::string errorMessage = "";
+
+      /**
        * Required.
        */
       bool required = false;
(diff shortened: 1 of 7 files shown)


linhaojun857:
I add a member variable `m_errorMessage` in class  `BaseObject::Property::Info` that user can customize error message for validating input, for convenience, I add a member variable `m_contextPtr` in class `Caret` that program can get user's customize error message and call `caret.setErrorMessage` to set it when meet `ERROR_INVALID_INTEGER`

linhaojun857:
@lganzzzo

linhaojun857:
@lganzzzo I have solved the code merge conflict

lganzzzo:
Hello @linhaojun857 ,

This PR is postponed due to the new functionality planned - see https://github.com/oatpp/oatpp/issues/879

Please feel free to comment on the proposal.

Issue #872, opened 2023-10-13
‘-fcf-protection=full’ is not supported on Raspberry Pi


alex-lt-kong:
I am trying to build Oatpp on Raspberry Pi 3 Model B Rev 1.2, my system's information is as follows:

```
Architecture:   aarch64
Distributor ID: Debian
Description:    Debian GNU/Linux 12 (bookworm)
Release:        12
Codename:       bookworm
```

Running `cmake ../` complains the following:
```
...
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
CMake Error at /usr/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Threads (missing: Threads_FOUND)
Call Stack (most recent call first):
  /usr/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.25/Modules/FindThreads.cmake:228 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/CMakeLists.txt:314 (find_package)


-- Configuring incomplete, errors occurred!
See also "/root/repos/oatpp/build/CMakeFiles/CMakeOutput.log".
See also "/root/repos/oatpp/build/CMakeFiles/CMakeError.log".
```

It says "Could NOT find Threads (missing: Threads_FOUND)" but this couldn't be the case as I am using `pthread` on the same device myself too (and I also use CMakeLists.txt, but I don't specify `find_package(Threads REQUIRED)` though).

After examining `/root/repos/oatpp/build/CMakeFiles/CMakeError.log`:

```
Performing C++ SOURCE FILE Test CMAKE_HAVE_LIBC_PTHREAD failed with the following output:
Change Dir: /root/repos/oatpp/build/CMakeFiles/CMakeScratch/TryCompile-YvNNHD

Run Build Command(s):/usr/bin/gmake -f Makefile cmTC_5e81e/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_5e81e.dir/build.make CMakeFiles/cmTC_5e81e.dir/build
gmake[1]: Entering directory '/root/repos/oatpp/build/CMakeFiles/CMakeScratch/TryCompile-YvNNHD'
[...]

PR #873, opened 2023-10-14
comment out fcf-protection

alex-lt-kong:
A minor patch per [issue 872](https://github.com/oatpp/oatpp/issues/872)

### DIFF
--- a/cmake/compiler-flags.cmake
+++ b/cmake/compiler-flags.cmake
@@ -253,7 +253,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
   add_compiler_flags(4.6 "-fstack-protector-strong")
   add_compiler_flags(4.6 "-grecord-gcc-switches")
   # Issue 872: https://github.com/oatpp/oatpp/issues/872
-  # -fcf-protection is supported on x86 GNU/Linux per this gcc doc:
+  # -fcf-protection is supported only on x86 GNU/Linux per this gcc doc:
   # https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection
   # add_compiler_flags(4.6 "-fcf-protection")
   add_compiler_flags(4.6 "-pipe")
--- a/cmake/compiler-flags.cmake
+++ b/cmake/compiler-flags.cmake
@@ -255,6 +255,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
   # Issue 872: https://github.com/oatpp/oatpp/issues/872
   # -fcf-protection is supported on x86 GNU/Linux per this gcc doc:
   # https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection
+  # add_compiler_flags(4.6 "-fcf-protection")
   add_compiler_flags(4.6 "-pipe")
   add_compiler_flags(4.6 "-Werror=format-security")
   add_compiler_flags(4.6 "-Wno-format-nonliteral")
--- a/cmake/compiler-flags.cmake
+++ b/cmake/compiler-flags.cmake
@@ -252,7 +252,9 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
   add_compiler_flags(4.6 "-fstack-clash-protection")
   add_compiler_flags(4.6 "-fstack-protector-strong")
   add_compiler_flags(4.6 "-grecord-gcc-switches")
-  add_compiler_flags(4.6 "-fcf-protection")
+  # Issue 872: https://github.com/oatpp/oatpp/issues/872
+  # -fcf-protection is supported on x86 GNU/Linux per this gcc doc:
+  # https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection
   add_compiler_flags(4.6 "-pipe")
   add_compiler_flags(4.6 "-Werror=format-security")
   add_compiler_flags(4.6 "-Wno-format-nonliteral")


lganzzzo (approved):

lganzzzo:
Merged!

cmake/compiler-flags.cmake
```diff
@@ -252,7 +252,10 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
   add_compiler_flags(4.6 "-fstack-clash-protection")
   add_compiler_flags(4.6 "-fstack-protector-strong")
   add_compiler_flags(4.6 "-grecord-gcc-switches")
-  add_compiler_flags(4.6 "-fcf-protection")
+  # Issue 872: https://github.com/oatpp/oatpp/issues/872
+  # -fcf-protection is supported only on x86 GNU/Linux per this gcc doc:
+  # https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection
+  # add_compiler_flags(4.6 "-fcf-protection")
```
fhuberts, line 258: This is fixed in the wrong way, now the hardening flag is not applied even when supported.
You should do (like line 306):
`add_compiler_flags(99 "-fcf-protection")
`

That tests the flags always

alex-lt-kong: @fhuberts : sorry for my ignorance. I just checked the `compiler-flags.cmake`file again. Given gcc's doc [here](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection), I propose replacing the current "fix" with the following:
```
if (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
    add_compiler_flags(4.6 "-fcf-protection")
endif ()
```
This solution is from this [SO answer](https://stackoverflow.com/a/26662907/19634193) and Cmake's doc on `CMAKE_SYSTEM_PROCESSOR` is [here](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html). Does it look better (or does it look good enough)?

fhuberts: Is this the right check?
Aren't you forgetting the check on Linux?
The linked GCC page  only mentions Linux but doesn't mention x64...

alex-lt-kong: Sorry I am very new to this platform-specific CMake thing... Regarding the Linux part, you are correct. On the x64 part, my own workstation's `CMAKE_SYSTEM_PROCESSOR` is `x86_64` and it has `-fcf-protection` support. But anyways, revised to the below:
```
if ((CMAKE_SYSTEM_NAME STREQUAL "Linux") AND (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)"))
    add_compiler_flags(4.6 "-fcf-protection")
endif ()
```
The Linux detection part is from this [SO answer](https://stackoverflow.com/a/70119658/19634193)

fhuberts: The GCC page doesn't say anything about it being for x64 only.
And on top of that it doesn't even say that this is Linux specific, only that it knows that Linux has an implementation for it.

That would mean that you only need the check on Linux (your proposal for that part is good).

alex-lt-kong: Sorry I dont quite get your comment "That would mean that you only need the check on Linux". Do you mean I can simplify the check to?:
```
if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
    add_compiler_flags(4.6 "-fcf-protection")
endif ()
```
Reading the gcc's doc again:

> Currently the **x86** **GNU/Linux** target provides an implementation based on Intel Control-flow Enforcement Technology (CET) which works for **i686 processor or newer**.

strictly speaking I think these three conditions must be met before we try to turn on `-fcf-protection`:

0. x86
1. GNU/Linux
2. i686 processor or newer

Also, the original PR is created because I tried to compile oat++ on my Raspberry Pi and it failed due to `-fcf-protection` being not supported on Raspberry Pi. My RPI's `CMAKE_SYSTEM_PROCESSOR` is `aarch64` and its `CMAKE_SYSTEM_NAME` is `Linux` (I am using the offcial Raspberry OS, which is a Debian-based distro). Given this, I think the `AND (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)")` part cannot be dropped (if dropping it is the intention of your latest comment)

fhuberts: Since you have practical proof, the 'Linux AND x64' check should do the trick. :-)

alex-lt-kong: Sorry does your 'Linux AND x64' mean I should include `"(x86)|(X86)"`:
```
if ((CMAKE_SYSTEM_NAME STREQUAL "Linux") AND (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)"))
    add_compiler_flags(4.6 "-fcf-protection")
endif ()
```
or I should include `"(x86)|(X86)|(amd64)|(AMD64)"` instead?:
```
if ((CMAKE_SYSTEM_NAME STREQUAL "Linux") AND (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)"))
    add_compiler_flags(4.6 "-fcf-protection")
endif ()
```

fhuberts: yeah I made a typo with x64, edited it to x86.
So I think it should be
```
if ((CMAKE_SYSTEM_NAME STREQUAL "Linux") AND (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)"))
    add_compiler_flags(4.6 "-fcf-protection")
endif ()
```

But for the review I defer to Leonid.
I'm merely the drive-by contributor that introduced the compiler flags stuff.

alex-lt-kong: okay, let me make a new PR later

PR #877, opened 2023-10-20
Prevent unintended removal of local CMake variables

sodevel:
Without quotes list expansion happens inside the set-call, if the list is empty, this actually unsets the variable. This is not intended and could lead to unwanted side effects. List expansion is not required there, quoting the variable results in setting the variable to empty for empty lists.


lganzzzo (approved):

lganzzzo:
Merged!

PR #878, opened 2023-10-20
Ignore more build directory patterns

sodevel:
This allows to use multiple build directories easily, the underscore pattern is for my very own selfish needs 😄. The removed patterns contain a typo and shouldn't be required for CMake builds anyway.


lganzzzo (approved):

lganzzzo:
Merged!

PR #883, opened 2023-10-29
New fix to issue 872

alex-lt-kong:
New fix of [872](https://github.com/oatpp/oatpp/issues/872) based on discussion with @fhuberts .

This fix tries to enable `-fcf-protection` when conditions are met, instead of simply disabling it all the time.

Tried building on a typical x86_64 Debian 12 (on which `-fcf-protection` is supported) and a Raspberry Pi (Raspberry Pi OS 12, on which `-fcf-protection` is not supported) and both are successful. `make test` on both devices are also successful.


fhuberts:
LGTM

lganzzzo (approved):

lganzzzo:
Merged!

PR #889, opened 2023-11-14
Allow overriding DI injected variables in ClientServerTestRunner

mwohlert:
Move OATPP_COMPONENT calls to constructor to allow supplying variables without using DI.

This is consistent with other constructors, which allow you to supply variables via the constructors instead of using the Oatpp environment DI.


lganzzzo (approved):

lganzzzo:
Looks good!
Merged!

PR #893, opened 2023-11-21
Fix compilation on GCC 4.6

### LINKED ISSUE #892: Build failure because of `-fcf-protection` (again)
Building on CentOS 7 fails because the compiler option `-fcf-protection` is wrongfully applied. The reason for this is quite simple, the test

https://github.com/oatpp/oatpp/blob/c1d37c189e3f07db3bd28d1ee98845692e094a06/cmake/compiler-flags.cmake#L258

does actually succeed for the detected architecture `x86_64`.

However, i don't know how to properly fix that. To build for the 32 bit architecture, you have to use a toolchain file and set the architecture manually. AFAIK for CentOS (and maybe others too) the proper name is `i686`, this won't pass the quoted test. To complicate the situation even more, you have two options to build for 32 bit: a true CMake crossbuild by specifying `CMAKE_SYSTEM_NAME` or just a regular build by using the command line option `-m32`. So far i was using the latter approach without issues, but this does not update `CMAKE_SYSTEM_PROCESSOR` and keeps whatever the system reports.

sodevel:
The compiler flag `-fcf-protection` is only available since GCC 8.1.

Closes #892


lganzzzo (approved):

lganzzzo:
Merged

PR #904, opened 2023-12-15
Fix the bug of send "null" in request body causes seg-fault

### LINKED ISSUE #902: Sending "null" in request body causes seg-fault on any endpoint w/ BODY_DTO
Observed that sending "null" request body to any endpoint w/ a body dto results in a segmentation fault. 

```
ENDPOINT("PUT", "/test", test. BODY_DTO(<Type>, body)) {
  return createResponse(Status::CODE_200, "OK");
}
``` 
Sending 
``` curl -X PUT "http://localhost:8000/test" -H  "accept: text/plain" -H  "Content-Type: application/json" -d "null" ```
causes segmentation fault.  Have reproduced with a variety of dto types and same behavior.

Suspect that root cause is related to the base ObjectMapper readFromString logic:
https://github.com/oatpp/oatpp/blob/57e6a714d757d3fcc85214bc59011f01bbdb1cc4/src/oatpp/core/data/mapping/ObjectMapper.hpp#L129-L131

A ParsingError is thrown whenever  the ObjectMapper fails to convert the json string in the request body to an object of the correct type.  However "null" is a valid json string and the json Deserializer code reflects this:

https://github.com/oatpp/oatpp/blob/57e6a714d757d3fcc85214bc59011f01bbdb1cc4/src/oatpp/parser/json/mapping/Deserializer.cpp#L182-L200

There is similar code for the other deserialization methods for different types.  The above suggests to me that the m_errorMessage in the caret instance passed to the deserializer is never initialized and remains set to nullptr.  Then when the ParsingError is constructed in the ObjectMapper, the ParsingError constructor attempts to dereference the nullptr to initialize the std::runtime_error instance and causes the segmentation fault:

https://github.com/oatpp/oatpp/blob/57e6a714d757d3fcc85214bc59011f01bbdb1cc4/src/oatpp/core/parser/ParsingError.cpp#L29-L34

I'm wondering if part of the fix is changing the check for whether to throw a parsing error in the ObjectMapper code to something like:

```
if(caret.getErrorMessage() != nullptr) {
   //throw
}
```
since there's not guaranteed to have been a parsing error just because no object was created.  Not sure whether the ObjectMapper base class extends to other syntaxes that don't contain explicit nulls where the existing implementation is correct.

linhaojun857:
fix #902

### DIFF
--- a/src/oatpp/core/data/mapping/ObjectMapper.hpp
+++ b/src/oatpp/core/data/mapping/ObjectMapper.hpp
@@ -126,7 +126,7 @@ public:
     auto type = Wrapper::Class::getType();
     oatpp::parser::Caret caret(str);
     auto result = read(caret, type).template cast<Wrapper>();
-    if(result == nullptr) {
+    if(caret.hasError()) {
       throw oatpp::parser::ParsingError(caret.getErrorMessage(), caret.getErrorCode(), caret.getPosition());
     }
     return result;
--- a/test/oatpp/web/FullTest.cpp
+++ b/test/oatpp/web/FullTest.cpp
@@ -342,6 +342,46 @@ void FullTest::onRun() {
       }
 
       {
+        String bodyIn = "null";
+        auto response = client->testBodyIsNull1(bodyIn, connection);
+        OATPP_ASSERT(response->getStatusCode() == 400)
+        auto returnedData = response->readBodyToString();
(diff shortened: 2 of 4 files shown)


lganzzzo (approved):

lganzzzo:
Merged!

PR #905, opened 2023-12-17
Fix the bug of request failed when multiple HTTP requests on the same connection and previous request not read body

linhaojun857:


Shu7i:
Hello,when can merge this pull request? I need this solution .

ns-wxin:
Is this bug going to be fixed at all? Or is it already fixed? if Yes, what version/build? @lganzzzo 
What was the reason this PR was not merged? @linhaojun857

PR #910, opened 2024-01-17
Client connectionprovider stop

dencrypt-jacob:
This solution works for me, but I am not sure if the code is correct for other scenarios.
So please let me know if changes are needed.


dencrypt-jacob:
I'll see if I can fix the failing testcases.

lganzzzo:
Hello @dencrypt-jacob ,

Thanks for the PR! Looks good!
I'll take a closer look later this week and I'll merge it.

Issue #912, opened 2024-01-25
How can I use OATPP to handle web browser request download file.


yifong0:
Hi All:
I surveyed the ticket list, and 2 tickets(#859,#638) talk about file download.
But I cannot get an answer from those tickets.
Does anyone know oatpp(web-server) how to handle remote web browser requests for a file download?
Thank you very much.
YiFong

lganzzzo:
Hello @yifong0 ,

Here is the simplest snippet:

```cpp
  ENDPOINT("GET", "file", getFile) {
    oatpp::String fileName = "my-file.txt"; // name of the downloaded file (how browser will see it)
    oatpp::String binaryData = "<data>"; // whatever binary data
    
    auto response = createResponse(Status::CODE_200, binaryData);
    response->putHeader("Content-Type", "application/octet-stream");
    response->putHeader("Content-Disposition", "attachment; filename=\"" + fileName + "\"");
    return response;
  }
```

yifong0:
Hi lganzzzo:

Your example is working, thank you very much.

The complete code for file download on my side.

URL = http://<your.server's ip>:<your.server's port>/<file_name>

ENDPOINT("GET", "/file/{file_name}", getFile, PATH(String, file_name)) {
    oatpp::String fileName = file_name; // name of the downloaded file (how browser will see it)
	oatpp::String fullfileName = "/usr/var/" + fileName;
    oatpp::String binaryData = oatpp::String::loadFromFile(fullfileName->c_str()); // whatever binary data
    auto response = createResponse(Status::CODE_200, binaryData);
    response->putHeader("Content-Type", "application/octet-stream");
    response->putHeader("Content-Disposition", "attachment; filename=\"" + fileName + "\"");
    return response;
  }

:D

Issue #859, opened 2023-09-20
Download a file from backend via GET method


ml16685:
Hello,

There is  the tutorial which describes how to upload a file to backend via POST method:
https://oatpp.io/docs/features/upload-file/

Is there Oatpp API to do the reverse operation, i.e. to download the file from backend via GET method? 
Maybe there is API description or an example?

Thanks in advance.

mrbesen:
> see [#715 (comment)](https://github.com/oatpp/oatpp/issues/715#issuecomment-1480438930)

I think that @ml16685 meant something different:
Imagine a process generates a dynamic file.
Now you want to provide a GET request for the user to access this file.
How would you do that? The only option i see is to load the file into an oatpp::String and use createResponse or
to implement a custom class implementing the Body abstract class.
I think an option to take an fd or file path and instruct oatpp to stream the content would be nice.

PR #958, opened 2024-07-28
FileBody to created responses with files

mrbesen:
This adresses #859 and #912.

Currently you have to load a file into a `oatpp::String` to use it as a response. This gets expensive fast if you have files that can be up to a few GB in size.
So i implemented a `FileBody` that creates the response from a memory mapped file. I am not familiar with Windows systemcalls when it comes to file access so i just implemented it using the simple `oatpp::String::loadFromFile` approach. 

Mapping the file into memory is very efficent on linux, as it allows the kernel to evict the memory pages on high memory pressure and repopulate them later. The memory of the mapped file can also be shared between processes reading the same file, reducing the memory pressure of the system even further.

### DIFF
--- a/src/oatpp/web/protocol/http/outgoing/FileBody.cpp
+++ b/src/oatpp/web/protocol/http/outgoing/FileBody.cpp
@@ -1,0 +1,125 @@
+/***************************************************************************
+ *
+ * Project         _____    __   ____   _      _
+ *                (  _  )  /__\ (_  _)_| |_  _| |_
+ *                 )(_)(  /(__)\  )( (_   _)(_   _)
+ *                (_____)(__)(__)(__)  |_|    |_|
+ *
+ *
+ * Copyright 2024-present, MrBesen
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


lganzzzo:
Hello @mrbesen ,

You can stream files or any other resource with existing API.
Use `StreamingBody` and `oatpp::data::resource::File`

### Simple API:

```cpp
  ENDPOINT("GET", "download", downloadFile) {
    auto body = std::make_shared<oatpp::web::protocol::http::outgoing::StreamingBody>(
       oatpp::data::resource::File("path/to/file").openInputStream()
    );
    return OutgoingResponse::createShared(Status::CODE_200, body);
  }
```

### Async API:

```cpp
  ENDPOINT_ASYNC("GET", "download", downloadFile) {
    
    ENDPOINT_ASYNC_INIT(downloadFile)
    
    Action act() override {
      auto body = std::make_shared<oatpp::web::protocol::http::outgoing::StreamingBody>(
        oatpp::data::resource::File("path/to/file").openInputStream()
      );
      return _return(OutgoingResponse::createShared(Status::CODE_200, body));
    }
    
  };
```

Regards,
Leonid

PR #1073, opened 2025-11-03
add `QueryResult::getRowsAffected()` to the API

ygerlach:
i know this breaks compatability with all database plugins.

Is there another way to expose this information?

Should it have an default implementation always returning 0? (To not break the compatability?)

### DIFF
--- a/src/oatpp/orm/QueryResult.cpp
+++ b/src/oatpp/orm/QueryResult.cpp
@@ -26,5 +26,8 @@
 
 namespace oatpp { namespace orm {
 
+  v_uint64 QueryResult::getRowsAffected() const {
+    throw std::runtime_error("[oatpp::orm::QueryResult::getRowsAffected()]: Error. Not implemented!");
+  }
 
 }}
--- a/src/oatpp/orm/QueryResult.hpp
+++ b/src/oatpp/orm/QueryResult.hpp
@@ -82,7 +82,7 @@ public:
    * Count of rows affected by the operation.
    * Might be 0
    */
-  virtual v_uint64 getRowsAffected() const = 0;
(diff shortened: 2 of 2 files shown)


lganzzzo:
Hello @ygerlach ,

Yes, it breaks the compatibility.

I think, that for such "custom" things it's better to go with a similar approach as in SQLite plug-in `oatpp::sqlite::Utils::getLastInsertRowId`. 
Basically it's just a util class that is able to read SQLite specific descriptor.
(You can find example usage [here](https://github.com/oatpp/example-crud/blob/325bc0720fd3079831f36989587bff6d6ee6541a/src/service/UserService.cpp#L9))

For postgres we can create a similar class::method `oatpp::postgres::Utils::getRowsAffected(result)`

Regards,
Leonid

ygerlach:
I am happy to modify my change to provide such a util class/function.
I always thought that the amount of affected rows is a normal thing (almost?) all DBMS return. Thats why i thought into adding it to the api.

lganzzzo:
> I always thought that the amount of affected rows is a normal thing (almost?) all DBMS return.

You are right, I've checked SQLite (I had doubts about it), PostgreSQL, and MySQL, and it looks that for all of them we have everything needed to make a call `QueryResult::getRowsAffected()` from QueryResult.

**Let's move with your original PR (this exact one) then:**

> Should it have an default implementation always returning 0? (To not break the compatability?)

Let's add a default implementation that just throws "Unimplemented"

lganzzzo (approved):

ygerlach:
@lganzzzo can you please review my implementation of this method for postgresql? https://github.com/oatpp/oatpp-postgresql/pull/34

PR #1086, opened 2026-01-29
Use nested namespace declarations

mikomikotaishi:
This PR simplifies namespace declarations, for example:
```cpp
// old-style
namespace oatpp { namespace web { namespace protocol { namespace http { namespace encoding {
    // ...
}}}}}

// new-style (C++17 onward)
namespace oatpp::web::protocol::http::encoding {
    // ...
}
```
Also, where parameters are `(void)`ed, the annotation `[[maybe_unused]]` is applied for API clarity.

As this syntax was introduced in C++17, there should not be any code breaking from this change. In fact this already existed in some parts of the code - this PR only applies it consistently.


mikomikotaishi:
The Windows build is failing because of an apparent lack of the flag `/std:c++17`, though this doesn't make much sense to me. The `CMakeLists.txt` itself contains the following
```cmake
set_target_properties(oatpp PROPERTIES
        CXX_STANDARD 17
        CXX_EXTENSIONS OFF
        CXX_STANDARD_REQUIRED ON
)
```
which should enable this. If anyone more familiar with this can provide clarification as to what is wrong, it would be much appreciated.

mikomikotaishi:
@lganzzzo Hi, could you review this PR?

